PR #24, opened 2021-09-14
Casminst 3049

cdelatte-hpe:
This will undo CASMREL-755 on the storage nodes so ceph can deploy a managed node-exporter.


upgrade/1.0/scripts/upgrade/ncn-upgrade-ceph-initial.sh
```diff
@@ -2,10 +2,6 @@
 #
 # Copyright 2021 Hewlett Packard Enterprise Development LP
 #
-set -e
```
denniswalker, line 5: Adding this back in per Craig's comment.

cdelatte-hpe: done

PR #33, opened 2021-09-15
0.9.6 Patch

SeanWallace:
First draft of 0.9.6 doc containing information mostly specific to the application of the Unbound fix.

### DIFF
--- a/upgrade/0.9/csm-0.9.6.md
+++ b/upgrade/0.9/csm-0.9.6.md
@@ -1,0 +1,164 @@
+# Non-Deterministic Unbound DNS Results Patch
+
+This procedure covers applying a new version of the `cray-dns-unbound` Helm chart to enable this setting in the configmap:
+
+```text
+rrset-roundrobin: no
+```
+
+Unbound back in April 2020 [changed](https://github.com/NLnetLabs/unbound/blob/master/doc/Changelog) the default of this setting to be `yes` which had the effect of randomizing the records returned from it if more than one entry corresponded (as woudl be the case for PTR records, for example):
+
+```text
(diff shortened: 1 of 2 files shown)


bo-quan (commented):
@SeanWallace Not sure if you're aware, we had a release/0.9 branch to hold the 0.9 upgrade instructions at https://github.com/Cray-HPE/docs-csm/blob/release/0.9/upgrade/0.9 . Does it make more sense to use release/0.9 branch for this commit?

SeanWallace:
> @SeanWallace Not sure if you're aware, we had a release/0.9 branch to hold the 0.9 upgrade instructions at https://github.com/Cray-HPE/docs-csm/blob/release/0.9/upgrade/0.9 . Does it make more sense to use release/0.9 branch for this commit?

Yep, good catch, that is the branch this should target. Thanks!

bo-quan:
To be consistent with previous two upgrades (0.9.4, and 0.9.5), I suggest to rename it as README.md under a subdir csm-0.9.6 instead.

SeanWallace:
> Thank you Sean for changing the release target. To be consistent with previous two upgrades (0.9.4, and 0.9.5), I suggest to rename it as README.md under a subdir csm-0.9.6 instead.

Well consistency seems to suggest that if there is only one README file and no other accompanying files then it's not under a subdir but rather under the main 0.9 folder with the full name. But it doesn't matter to me, I can move it if need be.

bo-quan:
There will be multiple changes for 0.9.6, so please move it under a directory to avoid refactoring in subsequent check-ins.

bklei (approved):

PR #54, opened 2021-09-20
Revert "CASMINST-3109: Update boot order set instructions"

denniswalker:
Reverts Cray-HPE/docs-csm#48


denniswalker:
Reverted the revert

PR #66, opened 2021-09-21
Add CASMREL-755 hotfix instructions to 0.9.6 patch instructions

bklei:


upgrade/0.9/csm-0.9.6/README.md
```diff
@@ -1,28 +1,35 @@
...
+   ```text
+   rrset-roundrobin: no
+   ```
 
-```text
-21 April 2020: George
-	- Change default value for 'rrset-roundrobin' to yes.
-	- Fix tests for new rrset-roundrobin default.
-```
+   Unbound back in April 2020 [changed](https://github.com/NLnetLabs/unbound/blob/master/doc/Changelog) the default of this setting to be `yes` which had the effect of randomizing the records returned from it if more than one entry corresponded (as woudl be the case for PTR records, for example):
```
kimjensen-hpe, line 12: woudl

kimjensen-hpe (approved):

PR #81, opened 2021-09-22
Firmware Docs Update - CASMTRIAGE-2275 / CASMTRIAGE-2293 / CAST-27463

mbuchmann-hpe:
Minor Updates to firmware documentation.

Updates for:
CASMTRIAGE-2275
CASMTRIAGE-2293

Moved FAS Loader Fails Doc to 0.9 branch


troubleshooting/known_issues/FAS_Loader_Fails.md
```diff
@@ -0,0 +1,67 @@
+## FAS LOADER FAILS
+
+**NOTE:** this procedure is only for csm-0.9.x releases.
```
nieuwsma, line 3: Why is this targeted for CSM 1.0? if its only for 0.9?

nieuwsma: Will this problem EVER occur on a 1.0 system where the loader is not a job but is part of the API?

nieuwsma (changes requested):

troubleshooting/known_issues/FAS_Loader_Fails.md
mbuchmann-hpe: Agree it should be in the release/0.9 branch - Will remove from this commit and add to 0.9/operations directory.
FAS Loader for 1.0 does rely on the repomd.xml file, so if that file is corrupted FAS Loader will not be able to read from Nexus.  The loader will not need the repomd.xml file to load a single .rpm file.

nieuwsma (approved):

denniswalker (approved):

PR #84, opened 2021-09-23
CASMPET-4896 - adding operational files needed for getting Reboot NCNs into 09x release branch

djlapenta:
…09 branch

Adding the files needed for getting an NCN Reboot procedure into the release/0.9 branch.


djlapenta:
just learned that I need to move the scripts dir to the root of csm-docs for release/0.9 branch..  not under operations..  I need to move that now and update all the references in the docs that I'm adding..

kimjensen-hpe (approved):

cdelatte-hpe:
Have we tested the scripts on a 0.9.x systems.   This is just a sanity question.  I can't see why they wouldn't work.

cdelatte-hpe (commented):
I reviewed the scripts as well and do not see anything that would not work on a 1.4 system.  If needed; these can be run but they are not all read only scripts  - add_pod_priority.sh  would make changes to many pods and might need to be reverted to get the test system back to the original state.  I'm good with merging these as is for 1.4

djlapenta:
@cdelatte-hpe,  the time that I believe that these scripts were definitely tested was as part of the CSM-095 patch testing.  I believe that Sean Wallace and Ryan Sjostrand ran through the entire end-to-end CSM-095 patch process with this document and all scripts (I put all of these same docs into the 095 patch and this incorporates clean-up changes that Sean and Ryan made to the reboot procedure when they were doing their testing).   That is the only case that I am aware of..  but yep, they were definitely tested then because the security patch required that all NCNs be rebooted.

PR #183, opened 2021-10-04
CASMINST-3037: Refine conman info in install/collecting_ncn_mac_addresses.md

dlaine-hpe:
Added details and example on conman usage to the documentation in the step where the user needs to connect to a console using conman.


install/collecting_ncn_mac_addresses.md
```diff
@@ -72,7 +72,22 @@ For help with either of those, see [LiveCD Setup](bootstrap_livecd_remote_iso.md
     pit# sleep 10
     pit# grep -oP "($mtoken|$stoken|$wtoken)" /etc/dnsmasq.d/statics.conf | sort -u | xargs -t -i ipmitool -I lanplus -U $USERNAME -E -H {} power on
     ```
-4. Now wait for the nodes to netboot. You can follow them with `conman -j ncn-*id*-mgmt` (use `conman -q` to see ). This takes less than 3 minutes, speed depends on how quickly your nodes POST.
+4. Now wait for the nodes to netboot. You can follow them with `conman -m ncn-*id*-mgmt` (use `conman -q` to see the list of nodes). This takes less than 3 minutes, speed depends on how quickly your nodes POST.  To exit the console session, use the command `&.` and to see a conman help screen for all supported escape sequences use `&?`.
```
haroldlongley, line 75: The use of "ncn-*id*-mgmt" is not clear.  We do not refer to "id" anywhere else in the documentation.   And our documentation style tries to avoid calling the node "m003" when it is really the hostname "ncn-m003".  The hostnames for the management nodes are of the form ncn-AXXX, where A is one of (m,s,w) and XXX is a 3 digit number with zero padding.  Don't need to explain this all here.

What I think you intend is that the hostname of the management node have the suffix "-mgmt" added to it.



Reorder to have this before the 2nd one
Use conman -q
Follow with conman -j

I notice you changed from conman -j to conman -m.  That makes sense for this use as read-only.  But you might want to highlight that this is readonly and mention that if there are problems and interaction with the node console is needed that they should switch from conman -m to conman -j.

haroldlongley (changes requested):

install/collecting_ncn_mac_addresses.md
```diff
@@ -50,7 +50,7 @@ For help with either of those, see [LiveCD Setup](bootstrap_livecd_remote_iso.md
     ```bash
     pit# mv /var/www/boot/script.ipxe /var/www/boot/script.ipxe.bak
     ```
-2. Verify consoles are active with `conman -q`,
+2. Verify consoles are active with `conman -q` - this lists all of the nodes that conman is connected to,
```
nrockershousen, line 53: Change to:

2. Verify consoles are active with `conman -q`.

The following command lists all nodes connected to ConMan.

install/collecting_ncn_mac_addresses.md
```diff
@@ -72,7 +72,27 @@ For help with either of those, see [LiveCD Setup](bootstrap_livecd_remote_iso.md
     pit# sleep 10
     pit# grep -oP "($mtoken|$stoken|$wtoken)" /etc/dnsmasq.d/statics.conf | sort -u | xargs -t -i ipmitool -I lanplus -U $USERNAME -E -H {} power on
     ```
-4. Now wait for the nodes to netboot. You can follow them with `conman -j ncn-*id*-mgmt` (use `conman -q` to see ). This takes less than 3 minutes, speed depends on how quickly your nodes POST.
+4. Now wait for the nodes to netboot. You can follow them with conman - the `-m` option follows the console output in read-only mode or the `-j` option joins an interactive console session. The available node names were listed in step 2 above.  This takes less than 3 minutes, speed depends on how quickly your nodes POST. To exit the console session, use the command `&.` and to see a conman help screen for all supported escape sequences use `&?`.
```
nrockershousen, line 75: - Remove "Now" from the start of this step.
- Instances of conman should be ConMan.
- Can we make a new step after this to exit the console session, and then remove that line from this block of text?

install/collecting_ncn_mac_addresses.md
dlaine-hpe: I struggled a little with the wording on this.  The list is not connected nodes, but the ones it is 'trying to connect to', or 'configured to connect to'.  It is entire possible to get a node on the list, call 'conman -j xname' and get a reply 'Not connected to xname'.  I don't know how confused people would be about that...

haroldlongley (approved):

nrockershousen (approved):

PR #204, opened 2021-10-05
Update grep command to pull out the test totals

jnowicki-hpe:


mharding-hpe (approved):
Probably should just delete this PR, though, since we are no longer maintaining the 1.1 docs.

rustydb (approved):

PR #220, opened 2021-10-06
CASMINST-3224: Document resetting CFS error count before NCN reboot (1.0)

rbak-hpe:
### Summary and Scope

Adds requested documentation on resetting the CFS error counts before rebooting NCNs.  Without this NCNs that had previously been in an error state would never leave that state or configure after reboot.

(release 1.2 changes this behavior and these steps will no longer be necessary)

### Issues and Related PRs

* Resolves CASMINST-3224

### Testing

Tested on:

* NA, docs change

### Risks and Mitigations

None

### DIFF
--- a/operations/CSM_product_management/Manage_a_Configuration_with_CFS.md
+++ b/operations/CSM_product_management/Manage_a_Configuration_with_CFS.md
@@ -26,7 +26,7 @@ management nodes in addition to the configuration required by CSM. Consult the
 documentation for these products for more information.
 
 For more detailed information on configuration management with CFS, see these
-topics in [Configuration Management](configuration_management/Configuration_Management.md)
+topics in [Configuration Management](../configuration_management/Configuration_Management.md)
    * Configuration Layers
    * Ansible Inventory
    * Configuration Management with the CFS Batcher
@@ -275,7 +275,7 @@ via NCN personalization.
 By default, the `csm.password` role reads passwords from Vault using the
 `secret/csm/management_nodes` secret and the `root_password` key in that secret.
 To set the password in Vault, follow steps 1-3 in the
-[Update NCN Passwords](#operations/security_and_authentication/Update_NCN_Passwords.md)
+[Update NCN Passwords](../security_and_authentication/Update_NCN_Passwords.md)
 procedure.
 
 This role is enabled by default in the CSM `site.yml` top-level play and assumes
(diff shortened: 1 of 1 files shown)


operations/node_management/Reboot_NCNs.md
```diff
@@ -138,6 +138,18 @@ The `kubectl` command is installed.
 
         If there are BGP Peering sessions that are not ESTABLISHED on either switch, refer to [Check BGP Status and Reset Sessions](../network/metallb_bgp/Check_BGP_Status_and_Reset_Sessions.md).
 
+    1. Check for components that are in an failed state in CFS.
```
mharding-hpe, line 141: Change 
```
are in an failed state
```
to
```
have `failed` status
```

operations/node_management/Reboot_NCNs.md
```diff
@@ -138,6 +138,18 @@ The `kubectl` command is installed.
...
+
+        Components can be found by running the following command:
```
mharding-hpe, line 143: Suggest changing to:
```
If there are any components with that status, this command will list them:
```

operations/node_management/Reboot_NCNs.md
```diff
@@ -138,6 +138,18 @@ The `kubectl` command is installed.
...
+        ```bash
+         ncn-m001# cray cfs components list --status failed
```
mharding-hpe, line 145: The beginning of the command prompt (the letter n) should be indented to the same depth as the three backticks that surround it

operations/node_management/Reboot_NCNs.md
```diff
@@ -138,6 +138,18 @@ The `kubectl` command is installed.
...
+        ```
+
+        For any NCN components found, reset the error count to 0.  Each component will also have to be disabled in CFS in order to not immediately trigger configuration.  The components will be re-enabled when they reboot.
+        ```bash
+         ncn-m001# cray cfs components update <xname> --error-count 0 --enabled false
```
mharding-hpe, line 150: The beginning of the command prompt (the letter n) should be indented to the same depth as the three backticks that surround it

operations/node_management/Reboot_NCNs.md
```diff
@@ -138,6 +138,18 @@ The `kubectl` command is installed.
...
         If there are BGP Peering sessions that are not ESTABLISHED on either switch, refer to [Check BGP Status and Reset Sessions](../network/metallb_bgp/Check_BGP_Status_and_Reset_Sessions.md).
 
+    1. Check for components that are in an failed state in CFS.
+
+        Components can be found by running the following command:
+        ```bash
+         ncn-m001# cray cfs components list --status failed
+        ```
+
+        For any NCN components found, reset the error count to 0.  Each component will also have to be disabled in CFS in order to not immediately trigger configuration.  The components will be re-enabled when they reboot.
```
mharding-hpe, line 148: There should only be single spacing between sentences.

Also suggest adding an additional sentence after this paragraph:
```text

**NOTE:** Be sure to replace the `<xname>` in the following command with the xname of the NCN component to be reset and disabled.
```

(people often get on autopilot in the docs, so we try to grab their attention when they cannot simply cut and paste a command as written)

PR #229, opened 2021-10-07
MTL-1482 NetworkBootRetries=0

rustydb:
If this value is not 0, and NetworkBootRetry=Disabled, then the
bios-baseline.sh script will error.


rustydb:
Re-opening with correct JIRA name #233

PR #230, opened 2021-10-07
CASMINST-3267: Add info on broken CFS feature to release notes (1.0)

rbak-hpe:
### Summary and Scope

Adds information on the broken "branch" parameters in CFS configurations to the known issues in the release notes.

### Issues and Related PRs

* Resolves CASMINST-3267
### Testing

Tested on:

* NA, docs only

### Risks and Mitigations

None

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,3 +4,4 @@
 ## Known Issues
 - Incorrect_output_for_bos_command_rerun: When a Boot Orchestration Service (BOS) session fails, it may output a message in the Boot Orchestration Agent (BOA) log associated with that session. This output contains a command that instructs the user how to re-run the failed session. It will only contain the nodes that failed during that session. The command is faulty, and this issue addresses correcting it.
 - Cfs_session_stuck_in_pending: Under some circumstances CFS sessions can get stuck in a pending state, never completing and potentially blocking other sessions.  This addresses cleaning up those sessions.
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -4,3 +4,4 @@
 ## Known Issues
 - Incorrect_output_for_bos_command_rerun: When a Boot Orchestration Service (BOS) session fails, it may output a message in the Boot Orchestration Agent (BOA) log associated with that session. This output contains a command that instructs the user how to re-run the failed session. It will only contain the nodes that failed during that session. The command is faulty, and this issue addresses correcting it.
 - Cfs_session_stuck_in_pending: Under some circumstances CFS sessions can get stuck in a pending state, never completing and potentially blocking other sessions.  This addresses cleaning up those sessions.
+- The "branch" parameter in CFS configurations may not work, and setting it will instead return an error.  Continue setting the commit id instead.
```
mharding-hpe, line 7: Instead of surrounding branch with double quotes, I recommend surrounding it with backticks -- that is how the rest of the docs format this type of thing (names of fields in objects, arguments to commands, command names themselves, Kubernetes pod names, etc).

Also, I think "commit id" should be "commit hash", technically. I was originally going to say it should be "commit ID", but I did some research and to my surprise, I cannot find anywhere in the git docs where they refer to it as an id, ID, or identifier. Usually they just refer to commits when they mean the hash, but when they do refer to the string itself, the docs use the term hash (at least as far as I discovered -- I by no means performed an exhaustive analysis).

PR #253, opened 2021-10-11
CASMINST-3285: Make upgrade more sophisticated about vault

mharding-hpe:
During the csm upgrade, vault is accessed in order to learn the BMC credentials. Currently it is hard-coded to look at vault pod 0 first, getting the username first, then the password. Only after that does it make sure it got a username, and if it did not, it then tries to get them both from vault pod 1. It then fails if it did not succeed in getting a username.

This PR changes this process in the following ways:
1) When getting the credentials, it checks to see if it got a username before it tries to get the password, since there's no sense in looking for the password if you failed to get the username.
2) Instead of always trying 0 and then 1, it now works from the list of Running vault pods, and works its way down the list, failing if none of them provide the needed info.

Fundamentally, though, it still does the same things. There were no changes to the underlying commands being run or the upgrade logic.

I tested the code (in isolation, not as part of an upgrade) on drax, ego, gamora, groot, lemondrop, loki, mug, odin, pepsi, rocket, shandy, slice, surtur, tab, and thanos. In all cases it worked as desired.


denniswalker:
Holding it back from the 1.0 release, but merging into 1.2.

PR #258, opened 2021-10-11
Restore Postgres for Capsules services (HAL-2379) (HAL-2380)

dmartinez31:
This is a duplicate of a PR that was previously in bitbucket. The changes are the same, but we had previously been waiting on machine resources to test which has now been completed.

Note: All of the instructions provide an example with "service-account" hyphenated which is not valid in postgres. The username actually gets automatically converted to "service_account" which is what all of the other services should update to in their  instructions as well.


dmartinez31:
@djlapenta and @kimjensen-hpe , I don't seem to be able to add reviewers here, but you two I think reviewed and approved the version of this that was on bitbucket. Not sure I understand the messages showing the jenkins check that has not completed. Is it just not getting build status reported back?

PR #355, opened 2021-10-20
CASMINST-2324: Clarify cabinets.yaml example for non-contiguous cabinets and VLAN IDs

haroldlongley:
CASMINST-2324: Clarify cabinets.yaml example for non-contiguous cabinets and VLAN IDs

### DIFF
--- a/install/create_cabinets_yaml.md
+++ b/install/create_cabinets_yaml.md
@@ -1,6 +1,6 @@
 # Create Cabinets YAML
 
-This page provides directions on constructing the `cabinets.yaml` file. This file lists cabinet IDs for any systems with non-contiguous cabinet ID numbers as well as VLAN overrides and controls how the `csi config init` command treats cabinet IDs.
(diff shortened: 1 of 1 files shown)


install/create_cabinets_yaml.md
```diff
@@ -20,14 +20,18 @@ cabinets:
...
+    - id: 1101
+      nmn-vlan: 2101
+      hmn-vlan: 3101
 - type: river
   total_number: 4
   starting_id: 3000
 ```
 
-In this example file, there are two Hill cabinets that will be automatically numbered as 9000 and 9001. The Mountain cabinets appear in three groupings of four IDs. The River cabinets are non-contiguous in four separated IDs.
+In this example file, there are two Hill cabinets that will be automatically numbered as 9000 and 9001. The Mountain cabinets appear in two groups as 1000 and 1001 and then another area in 1100 and 1101 with VLAN ids related to the cabinet number. This is one way to provide space in the cabinet IDs to allow for future expansion of the system or to identify location of a row of cabinets.  The River cabinets are contiguous in four separated IDs as 3000, 3001, 3002, and 30003..
```
nrockershousen, line 34: Change second sentence to:

The Mountain cabinets appear in two groups as 1000 and 1001, and then in another area as 1100 and 1101, with VLAN IDs related to the cabinet number.

PR #433, opened 2021-10-27
CASMCMS-7547: Add second boot set to the BOS session template example

rbak-hpe:
### Summary and Scope

Adds a second boot set to the BOS session template example to reinforce the idea that a template can have more than one boot set.

### Issues and Related PRs

* Resolves CASMCMS-7547

### Testing

Tested on:

* NA, docs change only

### Risks and Mitigations

None


rkleinman-hpe:
> You put a lot of effort into this one. Hey, it's concise, but still emphasizes the point.

Probably deserves some recognition points here.

nrockershousen (approved):

PR #520, opened 2021-11-08
CASMINST-3509: Correct wipe_ncn_disks_for_reinstallation

haroldlongley:
## Summary and Scope

CASMINST-3509: Correct wipe_ncn_disks_for_reinstallation
Several commands were added that had been used on the test on hela and the overall flow of the steps was adjusted to be more explicit upon which type of node the command should be run.

Also, indicate that install/create_ncn_metadata_csv.md will be pulling data from the SHCD HMN tab.


install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -228,34 +262,83 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
...
+      ```
 
-1. Stop the RAIDs.
+      Expected output when the etcd volume is present
 
-    ```bash
-    ncn# for md in /dev/md/*; do mdadm -S -v $md || echo nope ; done
-    ```
+      ```bash
+      ETCDLVM (254:1)
```
cdelatte-hpe, line 276: I don't think this will vary, but we should confirm it.  it could possibly be different depending on the hardware.  This is referring to the (254:1), not the ETCDLVM part of the output.   not a blocker.

cdelatte-hpe (approved):
1 comment, not a blocker, but we should confirm it.

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -178,14 +208,18 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
...
+        ncn-m# systemctl stop etcd.service
+        ncn-m# umount -v /run/lib-etcd /var/lib/etcd /var/lib/sdu
+
         ```
 
     2. Worker nodes.
 
         ```bash
-        ncn-w# umount -v /var/lib/containerd /var/lib/kubelet /var/lib/sdu
+        ncn-w# umount -v /run/containerd /run/lib-containerd /var/lib/containerd /var/lib/kubelet /var/lib/sdu
```
rustydb, line 222: This might've worked, but it isn't ideal. This is umounting the parts of the containerd overlayFS, ahead of the overlayFS itself. Should be reversed to ensure the overlayFS umounts first and can sync any work it has pending before we yank the underlying persistent storage mount.

Suggestion:
```
ncn-w# umount -v /var/lib/kubelet /var/lib/sdu /run/containerd /var/lib/containerd /run/lib-containerd
```

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -178,14 +208,18 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
 
     1. Master nodes.
 
+        Stop the etcd service on the master node before unmounting /var/lib/etcd
+
         ```bash
-        ncn-m# umount -v /var/lib/etcd /var/lib/sdu
+        ncn-m# systemctl stop etcd.service
+        ncn-m# umount -v /run/lib-etcd /var/lib/etcd /var/lib/sdu
```
rustydb, line 215: `umount` the overlayFS first to properly undo layering.

This might work as-is, but it would be proper to tell the kernel the overlayFS is going down first to remove doubt that the persistent storage was yanked out early.

```bash
ncn-m# umount -v /var/lib/etcd /run/lib-etcd /var/lib/sdu
```

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -95,80 +95,110 @@ wiping the disks and RAIDs.
...
-    1. Remove the VGs.
+       1. Remove the VGs.
 
-        ```bash
-        ncn-s# ls -1 /dev/sd* /dev/disk/by-label/*
-        ncn-s# vgremove -f -v --select 'vg_name=~ceph*'
-        ```
+           ```bash
+           ncn-s# ls -1 /dev/sd* /dev/disk/by-label/*
+           ncn-s# vgremove -f -v --select 'vg_name=~ceph*'
```
rustydb, line 200: These are cleaned up by dracut already, and are not subject to the wipe bug in MTL-1288.
This line could be removed.

For reference, the dracut code that removes these is here:
- Defined VGs to remove: https://github.com/Cray-HPE/dracut-metal-mdsquash/blob/main/90metalmdsquash/metal-md-lib.sh#L309-L310
- Code removing the VGs https://github.com/Cray-HPE/dracut-metal-mdsquash/blob/main/90metalmdsquash/metal-md-lib.sh#L325-L328

On the other hand, if this is desired still then MTL-1288 will remove these calls, @cdelatte-hpe might have an opinion here.

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -228,34 +262,83 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
...
+1. Remove etcd Volumes **on master nodes ONLY**.
+
+   ```bash
+   ncn-m# vgremove etcdvg0
+   ```
+
+1. Remove metal LVM.
+
+     ```bash
+     ncn# vgremove -f -v --select 'vg_name=~metal*'
```
rustydb, line 300: Same comment as line 200.

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -228,34 +262,83 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
...
+1. Group these commands together for each node.
+
+   This group of commands should be done in succession on one node before moving to do the same set of commands on the next node. The nodes would be addressed in descending order for each type of node.  Start with the utility storage nodes, then the worker nodes, then ncn-m003, then ncn-m002. 
+
+   > **WARNING:** Do not run these commands on ncn-m001
+
+   1. Stop the RAIDs.
+
+       ```bash
+       ncn# for md in /dev/md/*; do mdadm -S -v $md || echo nope ; done
```
rustydb, line 314: This should probably check status before and after.
```bash
cat /proc/mdstat
```

That would show any syncing RAIDs.

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -228,34 +262,83 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
...
+   1. List the disks for verification.
+
+       ```bash
+       ncn# ls -1 /dev/sd* /dev/disk/by-label/*
+       ```
+
+   1. Wipe the disks and RAIDs.
+
+       ```bash
+       ncn# wipefs --all --force /dev/sd* /dev/disk/by-label/*
```
rustydb, line 326: Does this need to be USB safe? This would fetch all non-USB busses that we use in Shasta:
```bash
doomed_disks=$(lsblk -l -o SIZE,NAME,TYPE,TRAN | grep -E '(raid|sata|nvme|sas)' | sort -u | awk '{print "/dev/"$2}' | tr '\n' ' ')
    for doomed_disk in $doomed_disks; do
        wipefs --all --force "$doomed_disk" 2> /dev/null
    done
```

(code borrowed from: https://github.com/Cray-HPE/dracut-metal-mdsquash/blob/main/90metalmdsquash/metal-md-lib.sh#L298)

install/wipe_ncn_disks_for_reinstallation.md
```diff
@@ -228,34 +262,83 @@ RAIDs, zeroing the disks, and then wiping the disks and RAIDs.
...
+       /dev/sdb: 6 bytes were erased at offset 0x00004000 (crypto_LUKS): 53 4b 55 4c ba be
+       /dev/sdc: 8 bytes were erased at offset 0x00000200 (gpt): 45 46 49 20 50 41 52 54
+       /dev/sdc: 8 bytes were erased at offset 0x6fc86d5e00 (gpt): 45 46 49 20 50 41 52 54
+       /dev/sdc: 2 bytes were erased at offset 0x000001fe (PMBR): 55 aa
+       ```
 
-   See [Basic Wipe](#basic-wipe) section for expected output from the `wipefs` command.
+       Verify there are no error messages in the output.
 
+       The `wipefs` command may fail if no labeled disks are found, which is an indication of a larger problem.
```
rustydb, line 344: The comment I left on +326 would mitigate false-positives on disks, but not missing labels.

rustydb (changes requested):

install/wipe_ncn_disks_for_reinstallation.md
rustydb: In hindsight, this is just wiping the magic bit.

The order doesn't matter, if this was actually erasing or writing zeros then I'd say the order would matter.

This can be resolved.

install/wipe_ncn_disks_for_reinstallation.md
haroldlongley: Not changing this until there is more feedback.

install/wipe_ncn_disks_for_reinstallation.md
haroldlongley: We already have a warning not to run this command on ncn-m001 which would be the only node with a USB device, unless the customer has done something non-standard.

install/wipe_ncn_disks_for_reinstallation.md
haroldlongley: Not sure what change you are requesting.

install/wipe_ncn_disks_for_reinstallation.md
haroldlongley: Changed order.

install/wipe_ncn_disks_for_reinstallation.md
haroldlongley: Added status check before and after.

install/wipe_ncn_disks_for_reinstallation.md
rustydb: @haroldlongley i think we can leave this as is.

I'll pencil this into my notes for MTL-1288.

I'll leave this open for you to resolve, incase Craig wants to comment.

rustydb (approved):

cdelatte-hpe (approved):

haroldlongley:
Made new PRs for 1.2 and main

PR #703, opened 2022-01-06
CASMINST-3509 PR 1.2 Correct wipe_ncn_disks_for_reinstallation

haroldlongley:
## Summary and Scope

CASMINST-3509: Correct wipe_ncn_disks_for_reinstallation
Several commands were added that had been used on the test on hela and the overall flow of the steps was adjusted to be more explicit upon which type of node the command should be run.

Also, indicate that install/create_ncn_metadata_csv.md will be pulling data from the SHCD HMN tab.


haroldlongley:
Backport to main failed, so made separate PR to main.

PR #917, opened 2022-02-16
CASMINST-4039: Clarify power off preparation procedure

jack-stanek-hpe:
This change improves the "Prepare the System for Power Off" procedure
to clarify how SAT is used in this process

  - Additional context was added to the "Prerequisites" section to
    indicate the reason for linking the user to other pages and
    documents.
  - The order of steps 2 and 3 was reversed, as a SAT authentication
    token may be necessary for using `sat status`.
  - Additional context was added to step 2 (previously step 3) in order
    to indicate the reason for linking the user to other pages and
    documents. The text of this change is identical to that of the
    change noted above.
  - The link to the "Authenticate an Account with the Command Line" page
    was removed due to it not being relevant to step 2 (previously step
    3.)

### DIFF
--- a/operations/power_management/Prepare_the_System_for_Power_Off.md
+++ b/operations/power_management/Prepare_the_System_for_Power_Off.md
@@ -8,7 +8,7 @@ The `sat bootsys shutdown` and `sat bootsys boot` commands are used to shut down
 
 ### Prerequisites
 
-An authentication token is required to access the API gateway and to use the `sat` command. See the [System Security and Authentication](../security_and_authentication/System_Security_and_Authentication.md) and "SAT Authentication" in the System Admin Toolkit (SAT) product stream documentation.
+An authentication token is required to access the API gateway and to use the `sat` command. See the "SAT Authentication" section of the HPE Cray EX System Admin Toolkit (SAT) product stream documentation (S-8031) for instructions on how to acquire a SAT authentication token.
 
 ### Procedure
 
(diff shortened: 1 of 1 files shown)


operations/power_management/Prepare_the_System_for_Power_Off.md
```diff
@@ -37,7 +37,13 @@ An authentication token is required to access the API gateway and to use the `sa
...
+2.  Use `sat auth` to authenticate to the API gateway within SAT.
+
+    If SAT has already been authenticated to the API gateway, this step may be skipped.
+
+    See the "SAT Authentication" section in the System Admin Toolkit (SAT) product stream documentation (S-8031) for instructions on how to acquire a SAT authentication token.
+
+3.  Determine which Boot Orchestration Service \(BOS\) templates to use to shut down compute nodes and UANs. You can list all the session templates using `cray bos v1 sessiontemplate list`. If you are unsure of which template is in use, you can call `sat status` to find the xname, then use `cray cfs components describe XNAME` to find the bos_session, and use `cray bos v1 session describe BOS_SESSION` to find the `templateUuid`. Then finally use `cray bos v1 sessiontemplate describe TEMPLATE_UUID` to determine the list of xnames associated with a given template. For example:
```
nrockershousen, line 46: Instead of listing all of the steps in a block of text, it would be better to break this down into sub-steps. Something like the following could work better:

3. Determine which Boot Orchestration Service \(BOS\) templates to use to shut down compute nodes and UANs.
    
    There will be a seperate session template for UANs and computes nodes.
    
    1. List all the session templates.
        
        If it is unclear what session template is in use, proceed to the next substep.

        ```
        ncn-m001# cray bos v1 sessiontemplate list
        ```

    2. Find the xname with `sat status`.

        ```bash
        ncn-m001# sat status | grep "Compute\|Application"
        ```

        Example output:

        ```
        | x3000c0s19b1n0 | Node | 1        | On    | OK   | True    | X86  | River | Compute     | Sling    |
        | x3000c0s19b2n0 | Node | 2        | On    | OK   | True    | X86  | River | Compute     | Sling    |
        | x3000c0s19b3n0 | Node | 3        | On    | OK   | True    | X86  | River | Compute     | Sling    |
        | x3000c0s19b4n0 | Node | 4        | On    | OK   | True    | X86  | River | Compute     | Sling    |
        | x3000c0s27b0n0 | Node | 49169248 | On    | OK   | True    | X86  | River | Application | Sling    |
        ```

    3. Find the `bos_session` value via the Configuration Framework Service (CFS).

        ```
        ncn-m001# cray cfs components describe XNAME | grep bos_session
        ```

        Example output:

        ```
        bos_session = "e98cdc5d-3f2d-4fc8-a6e4-1d301d37f52f"
        ```

    4. Find the required `templateUuid` value with BOS.
        
        ```
        ncn-m001# cray bos v1 session describe XNAME | grep templateUuid
        ```

        Example output:

        ```
        templateUuid = "compute-nid1-4-sessiontemplate"
        ```

    5. Determine the list of xnames associated with the desired boot session template.

        ```
        ncn-m001# cray bos v1 sessiontemplate describe SESSION_TEMPLATE_NAME | grep node_list
        ```

        Example output:

        ```
        node_list = [ "x3000c0s19b1n0", "x3000c0s19b2n0", "x3000c0s19b3n0", "x3000c0s19b4n0",]
        ```

nrockershousen (approved):

operations/power_management/Prepare_the_System_for_Power_Off.md
```diff
@@ -8,7 +8,7 @@ The `sat bootsys shutdown` and `sat bootsys boot` commands are used to shut down
 
 ### Prerequisites
 
-An authentication token is required to access the API gateway and to use the `sat` command. See the [System Security and Authentication](../security_and_authentication/System_Security_and_Authentication.md) and "SAT Authentication" in the System Admin Toolkit (SAT) product stream documentation.
+An authentication token is required to access the API gateway and to use the `sat` command. See the [System Security and Authentication](../security_and_authentication/System_Security_and_Authentication.md) section of this guide for information on how authentication tokens are used, and the "SAT Authentication" section of the HPE Cray EX System Admin Toolkit (SAT) product stream documentation (S-8031) for instructions on how to acquire a SAT authentication token.
```
rgross727, line 11: The page that the "System Security and Authentication" link here goes to was changed some time ago and no longer has the information that you are looking for. @nrockershousen may have a perspective on what the link here should be.

jack-stanek-hpe: My initial impression was that the `System Security and Authentication` page was linked in order to justify why the authentication step was necessary (i.e. the token is needed for API requests SAT makes.) Thus that plus the SAT product documentation combined would be sufficient to explain why and how to acquire a JWT for API calls. Still, @nrockershousen I'm also curious if there's a better page to link to here, please let me know.

nrockershousen: The actual procedure to retrieve an authentication token is: **operations/security_and_authentication/Retrieve_an_Authentication_Token.md**

The current link in the PR is just an overview of security and authentication in general. If you don't want to link to the actual procedure and want a list of all topics useful for authentication, I recommend linking to the security and authentication section the in operations index file:
**operations/index.md#security-and-authentication**

jack-stanek-hpe: I'm a little unsure about linking to the page on retrieving an authentication token by hand, as it doesn't specify how to use the retrieved token with SAT (and, as an aside, nor should it.) SAT has its own authentication mechanism through `sat auth` which essentially automates the documented authentication flow for SAT.

Linking to the Security and Authentication section index might be more appropriate in that case, though it may be confusing to the user as most of the subsections are not really relevant to retrieving an auth token for SAT.

My current leaning given the above is that it might be best to just remove the link entirely, and direct the user to the SAT product stream documentation exclusively.

Issue #952, opened 2022-02-23
Authoring of triple-backtick codblocks prevents rendering in Jekyll/Liquid


pawsey-kbuckley:
There are a swathe of instances, throughout the source of these docs,
where "code blocks" are begun with the triple-backtick marker, but
are never closed with a corresponding one.

Whilst appreciating that you can "get away with it" in some rendering
environments, it's rather poor form, and,  more importantly,  it does 
break rendering in some envirtonments, with Jekyll/Liquid being the
one I have been trying to use locally.

Given that there are examples of code block markup where the author
has done the "right thing", it's clearly not some  "internal style guide 
authoring thing": more likely the result of an ad-hoc approach?

Any chance of fixing this up? 

Reason for not forking is that there is just too much of it to clean up!

denniswalker:
Appreciate the feedback. We're looking into this now.

gbaker-hpe:
Thanks for creating the issue and for the follow-up PRs.  This is a good change and can merge once the PRs are fixed up.

pawsey-kbuckley:
Just following up on this, now the heat is off a bit here.

I gave up trying to re-base and just cloned the latest main branch
and have just pulled it up to date, so as to be at commit

```
3098cced5a0 (upstream/main, upstream/HEAD, main) CASMINST-4695: More linting
```

The good news:

I have only found one instance of triple-backticks code-blocks not having 
been coded correctly.

That's at line 539 in this file,

```
./operations/kubernetes/Troubleshoot_Postgres_Database.md
```
where there's a block closed with a double tick, not a triple.


Some other news:

There's one place where the basic script I have been running to check, for
unclosed triple-backticks running into other Markdown commands, trips up,
and that's because of the use of a double-hash inside a code block, eg

```
## This example assumes the csm-1.0.0 release is currently running and the csm-1.0.0.tar.gz has been pulled down under /root
ncn# cd /root
ncn# tar -xzf csm-1.0.0.tar.gz
```

where the double-hash is the 2nd-level heading markup in Markdown.

FWIW, that's at line 73 in 

```
./operations/CSM_product_management/Post_Install_Customizations.md
```

and it's worth noting that the comment pretty much echoes the text in the
paragraph above the code-block anyway.

Now yes, I could try and write a better parser, but do you (or perhaps: do the docs) need 
to have comments inside the code-blocks anyway?

It doesn't look, given the prompts, as though such code-blocks are meant to
[...]

PR #978, opened 2022-02-28
Authoring of triple-backtick codblocks prevents rendering in Jekyll/Liquid (Follow up)

pawsey-kbuckley:
## Summary and Scope

In my original reporting of the underlying issue

https://github.com/Cray-HPE/docs-csm/issues/952

I'd stated that

> Reason for not forking is that there is just too much of it to clean up!

although, in absence of any feedback around that issue, I decided to
take a closer look at the problem and, after some head scratching, as
regards trying to see where triple-backtick code blocks have not been
properly closed off, have narrowed down a smaller set of instances, and, 
one instance, which turned out to be the one that made it appear that 
there were many more issues than now seems to be the case, where
triple-backticks had been used "in-line", when there is no need to use
them.

Hope this patch helps you: it has certainly helped me!


mharding-hpe (changes requested):
This needs to be rebased and have its conflicts resolved in order for it to be reviewed.

Also, there is a more recent PR that may cover most or all of these concerns: https://github.com/Cray-HPE/docs-csm/pull/1310

PR #1010, opened 2022-03-02
Add a note about connecting to the NCNs while updating the SSH Keys.

jnowicki-hpe:
## Summary and Scope

Add a note to indicate while running the prerequisites.sh script, during the UPDATE_SSH_KEYS phase, questions about connecting will be needed. This is a documentation change to clarify this part of this process.

## Issues and Related PRs

* Resolves [CASMINST-4092](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-4092)

## Testing

Documentation change


mharding-hpe (changes requested):
This needs to be rebased and have its conflicts resolved in order for it to be reviewed.

upgrade/1.2/Stage_0_Prerequisites.md
```diff
@@ -171,6 +171,14 @@ Run check script:
    ncn-m001# /usr/share/doc/csm/upgrade/1.2/scripts/upgrade/prerequisites.sh --csm-version [CSM_RELEASE] --tarball-file [PATH_TO_CSM_TARBALL_FILE]
    ```
 
+**NOTE** During the UPDATE\_SSH\_KEYS phase, you will need to respond to interactive prompts for connecting to each NCN. Respond with yes to these prompts. Here is an example:
+
+```bash
```
rustydb, line 176: This should probably change to `text` or `console` since this isn't technically bash and it shouldn't get code-coloring/highlighting like a bash script would.

rustydb (changes requested):

mharding-hpe (approved):

mharding-hpe:
Actually, closing this PR now, per this comment in the ticket:
> After reviewing this ticket and the current documentation, I think it would be best to not add an additional note. This is a 
> fairly common item which I think it is reasonable for even someone not experienced with installs/upgrades to recognize and 
> know how to handle. It seems like this would just add more clutter to the documentation without adding much value.

PR #1202, opened 2022-03-30
CASMINST-4385: Remove upgrade documents and scripts for csm-1.2 and earlier

mharding-hpe:
It has not yet been determined what purpose the docs-csm main branch serves, but there does appear to be agreement that nothing pre-1.3 needs to be in there. This PR removes the 1.2 and earlier upgrade directories from the main branch, to reduce confusion in this regard.


dzou-hpe (approved):
thanks @mharding-hpe

gbaker-hpe (approved):

PR #1320, opened 2022-04-12
CASMINST-4443 ignore another expected error

dzou-hpe:
## Summary and Scope

ignore another expected error found during upgrade

_Is this change backwards incompatible, backwards compatible, or a backwards compatible bugfix?_

## Issues and Related PRs

* Resolves [issue id](issue link)
* Change will also be needed in `<insert branch name here>`
* Future work required by [issue id](issue link)
* Documentation changes required in [issue id](issue link)
* Merge with/before/after `<insert PR URL here>`

## Testing

_List the environments in which these changes were tested._

### Tested on:

  * `<development system>`
  * Local development environment
  * Virtual Shasta

### Test description:

_How were the changes tested and success verified? If schema changes were part of this change, how were those handled in your upgrade/downgrade testing?_

- Were the install/upgrade-based validation checks/tests run (goss tests/install-validation doc)?
- Were continuous integration tests run? If not, why?
- Was upgrade tested? If not, why?
- Was downgrade tested? If not, why?
- Were new tests (or test issues/Jiras) created for this change?

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_

## Pull Request Checklist

- [ ] Version number(s) incremented, if applicable
- [ ] Copyrights updated
- [ ] License file intact
- [ ] Target branch correct
- [ ] CHANGELOG.md updated
- [ ] Testing is appropriate and complete, if applicable
- [ ] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable

### DIFF
--- a/upgrade/1.2/scripts/common/upgrade-state.sh
+++ b/upgrade/1.2/scripts/common/upgrade-state.sh
@@ -98,6 +98,12 @@ function err_report() {
     if [[ ${shouldIgnore} -eq 1 ]]; then
         return 0
     fi
+
+    ignoreCmd="grep -v -e Running -e Completed > /dev/null"
+    shouldIgnore=$(echo "$cmd" | grep "${ignoreCmd}" | wc -l)
+    if [[ ${shouldIgnore} -eq 1 ]]; then
+        return 0
+    fi
     
     # force output to console regardless of redirection
     echo >/dev/tty 
--- a/upgrade/1.2/scripts/common/upgrade-state.sh
+++ b/upgrade/1.2/scripts/common/upgrade-state.sh
@@ -98,12 +98,6 @@ function err_report() {
     if [[ ${shouldIgnore} -eq 1 ]]; then
         return 0
     fi
-
-    ignoreCmd="grep -v -e Running -e Completed > /dev/null"
-    shouldIgnore=$(echo "$cmd" | grep "${ignoreCmd}" | wc -l)
-    if [[ ${shouldIgnore} -eq 1 ]]; then
-        return 0
-    fi
     
     # force output to console regardless of redirection
     echo >/dev/tty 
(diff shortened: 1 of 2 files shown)


djlapenta (approved):

johren-hpe (changes requested):
Where/how was this tested?

dzou-hpe:
> Where/how was this tested?

this is not reproducible so i basically had to assume this will work like similar "fixes" we had done before

upgrade/1.2/scripts/common/upgrade-state.sh
```diff
@@ -98,6 +98,12 @@ function err_report() {
     if [[ ${shouldIgnore} -eq 1 ]]; then
         return 0
     fi
+
+    ignoreCmd="grep -v -e Running -e Completed > /dev/null"
```
johren-hpe, line 102: I'm not following how we think that line is tripping an error because there is a `set +e` and `set -e` around that set of lines.    Could you explain more?

johren-hpe (changes requested):

upgrade/1.2/scripts/common/upgrade-state.sh
dzou-hpe: our set +e will make sure it doesn't exist but it is caught by our `trap 'err_report' ERR`. and this error is coming from our trap function

johren-hpe: Okay, I see that know.   It seems like there is a better way to handle this.  Looking.

johren-hpe: Okay, how about 
```
echo "$output" | grep -v -e Running -e Completed > /dev/null || rc=$?
```
instead of 
```
echo "$output" | grep -v -e Running -e Completed > /dev/null
rc=$?
```

johren-hpe: That will make the command always have a non-zero return code but still set the rc for the grep part of the command.

dzou-hpe: that is better.

johren-hpe (approved):

PR #1328, opened 2022-04-13
MTL-1647: adjust docs to account for no default root password/keys

mharding-hpe:
@heemstra made a PR for MTL-1647 a little while ago which updated the docs to cover the fact that in csm-1.2 the NCN images no longer have a default password or SSH keys. Some of the changes in that PR were reverted, presumably on accident. This PR restores the lost changes.

### DIFF
--- a/install/deploy_management_nodes.md
+++ b/install/deploy_management_nodes.md
@@ -264,7 +264,7 @@ The configuration workflow described here is intended to help understand the exp
 <a name="deploy"></a>
 ### 3.2 Deploy
 
-1. Set the default root password and SSH keys and optionally change the timezone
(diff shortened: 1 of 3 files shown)


install/deploy_management_nodes.md
```diff
@@ -264,13 +264,11 @@ The configuration workflow described here is intended to help understand the exp
 <a name="deploy"></a>
 ### 3.2 Deploy
 
-1. Change the default root password and SSH keys
+1. Set the default root password and SSH keys and optionally change the timezone
```
cdelatte-hpe, line 267:
```suggestion
1. Set the default root password and SSH keys and optionally change the timezone.
```

cdelatte-hpe (approved):

PR #1380, opened 2022-04-19
Added Graffana documentation

mantoanibroadwing:
Adding Grafana Documentation

### DIFF
--- a/troubleshooting/grafana.md
+++ b/troubleshooting/grafana.md
@@ -1,0 +1,37 @@
+#Grafana Links
+
+This document provides links to all environment with links to its Grafana.
+Default access is possible using `vers` ID and password.
+
+| Environment | Grafana Link                            |
+| ----------- | --------------------------------------- |
+| baldar      | https://grafana.baldar.dev.cray.com     |
+| drax        | https://grafana.drax.dev.cray.com       |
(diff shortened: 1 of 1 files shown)


djlapenta (changes requested):
Hi Rodrigo,  I will update the ticket as well, but I think the goal for adding grafana information to troubleshooting was to get a list of dashboards that we have available in grafana that might help those needing to do troubleshooting..  i.e. for Ceph, the following dashboards are available;  For Networking, the following dashboards are available - something like that.    The troubleshooting docs within docs-csm are visible to the customer so it does not seem like the appropriate place to list out the URLs for the internal systems.   I think it would be more useful to outline the list of dashboards that are available for use, for the troubleshooting section of the documentation.

mharding-hpe (changes requested):
We do not want links to internal systems in customer-facing docs.

mharding-hpe:
I will try to shepherd this through to completion.

mharding-hpe:
I think this PR should be closed. The desired content is covered by these PRs:
1.2: https://github.com/Cray-HPE/docs-csm/pull/1434
main: https://github.com/Cray-HPE/docs-csm/pull/1858

That just needs to get backported to 1.0 in place of this PR. So I will do that.

PR #1381, opened 2022-04-19
Added index.md documentation with all troubleshoot links

mantoanibroadwing:
Fixed links and populated index.md file


johren-hpe (changes requested):
Same comments as the PR for the 1.2 branch.

johren-hpe:
Most of the links are still absolute paths. These all should be relative paths.

I suspect it also has the same dead link as the one in the 1.2 branch.  We just don't run the mardown-link-check in this branch.

PR #1432, opened 2022-04-22
CASMINST-4502 Ensure PITDATA is set

rustydb:
## Summary and Scope

_Summarize what has changed. Explain why this PR is necessary. What is impacted? Is this a new feature, critical bug fix, etc?_

Ensure that the `umount` call in the USB directions uses the absolute path to the disk-label, since the PITDATA env var is no longer set. Also remove hardcodes for setting PITDATA to `/var/www/ephemeral` and set it
via `lsblk`.

_Is this change backwards incompatible, backwards compatible, or a backwards compatible bugfix?_

No.

## Issues and Related PRs

* Resolves [CASMINST-4502](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-4502)
* Change will also be needed in `release/1.2`

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_

## Pull Request Checklist

- [ ] Version number(s) incremented, if applicable
- [ ] Copyrights updated
- [ ] License file intact
- [ ] Target branch correct
- [ ] CHANGELOG.md updated
- [ ] Testing is appropriate and complete, if applicable
- [ ] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable

### DIFF
--- a/install/bootstrap_livecd_remote_iso.md
+++ b/install/bootstrap_livecd_remote_iso.md
@@ -135,6 +135,36 @@ On first login (over SSH or at local console) the LiveCD will prompt the adminis
    = PIT Identification = COPY/CUT END =========================================
    ```
 
+1. Find a local disk for storing product installers.
+
+   ```bash
+   pit# disk="$(lsblk -l -o SIZE,NAME,TYPE,TRAN | grep -E '(sata|nvme|sas)' | sort -h | awk '{print $2}' | head -n 1 | tr -d '\n')"
+   pit# echo $disk
+   pit# parted --wipesignatures -m --align=opt --ignore-busy -s /dev/$disk -- mklabel gpt mkpart primary ext4 2048s 100%
+   pit# mkfs.ext4 -L PITDATA "/dev/${disk}1"
+   pit# mount -vL PITDATA
+   ```
+
+   The `parted` command may give an error similar to the following:
+   ```text
+   Error: Partition(s) 4 on /dev/sda have been written, but we have been unable to inform the kernel of the change, probably
+   because it/they are in use. As a result, the old partition(s) will remain in use. You should reboot now before making
+   further changes.
+   ```
+
+   In that case, the following steps may resolve the problem without needing to reboot. These commands remove
+   volume groups and raid arrays that may be using the disk. **These commands only need to be run if the earlier
+   `parted` command failed.**
+
+   ```bash
+   pit# RAIDS=$(grep "${disk}[0-9]" /proc/mdstat | awk '{ print "/dev/"$1 }') ; echo ${RAIDS}
+   pit# VGS=$(echo ${RAIDS} | xargs -r pvs --noheadings -o vg_name 2>/dev/null) ; echo ${VGS}
+   pit# echo ${VGS} | xargs -r -t -n 1 vgremove -f -v
+   pit# echo ${RAIDS} | xargs -r -t -n 1 mdadm -S -f -v
+   ```
+
+   After running the above procedure, retry the `parted` command which failed. If it succeeds, resume the install from that point.
+
 1. <a name="set-up-site-link"></a>Set up the site-link, enabling SSH to work. You can reconnect with SSH after this step.
    > **Note:** If your site's network authority or network administrator has already provisioned a DHCP IPv4 address for your master node's external NIC(s), **then skip this step**.
 
@@ -190,6 +220,7 @@ On first login (over SSH or at local console) the LiveCD will prompt the adminis
          ```bash
          pit# CSM_RELEASE=csm-x.y.z
          pit# SYSTEM_NAME=eniac
+         pit# PITDATA=$(lsblk -o MOUNTPOINT -nr /dev/disk/by-label/PITDATA)
          ```
 
       1. Add variables to the PIT environment.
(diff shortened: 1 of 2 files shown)


install/bootstrap_livecd_remote_iso.md
```diff
@@ -202,9 +232,9 @@ On first login (over SSH or at local console) the LiveCD will prompt the adminis
 
          ```bash
          pit# echo "
-         CSM_PATH=/var/www/ephemeral/${CSM_RELEASE}
          CSM_RELEASE=${CSM_RELEASE}
-         PITDATA=/var/www/ephemeral
+         PITDATA=$(lsblk -o MOUNTPOINT -nr /dev/disk/by-label/PITDATA)
+         CSM_PATH=${PITDATA}/${CSM_RELEASE}
```
mharding-hpe, line 237: This won't work because the PITDATA variable isn't set. The line above it doesn't set it -- it just echoes the assignment to be written to the /etc/environment file. If you want to use the $PITDATA variable within this echo, it will need to be first set before this echo is done.

install/bootstrap_livecd_usb.md
```diff
@@ -517,9 +517,9 @@ Now that the configuration is generated, the LiveCD must be populated with the g
 
    ```bash
    linux# echo "
-   CSM_PATH=/var/www/ephemeral/${CSM_RELEASE}
    CSM_RELEASE=${CSM_RELEASE}
-   PITDATA=/var/www/ephemeral
+   PITDATA=$(lsblk -o MOUNTPOINT -nr /dev/disk/by-label/PITDATA)
```
mharding-hpe, line 521: That lsblk command will work and give us /var/www/ephemeral even though we aren't yet booted from the PIT, and PITDATA is still mounted over /mnt/pitdata? Since this is just being written into the environment file of the PIT node, we want it to have the correct value for the booted PIT, not its current value.

install/bootstrap_livecd_usb.md
```diff
@@ -517,9 +517,9 @@ Now that the configuration is generated, the LiveCD must be populated with the g
...
+   CSM_PATH=${PITDATA}/${CSM_RELEASE}
```
mharding-hpe, line 522: This assignment has the same problem as the one in the remote ISO document -- this is not going to use the PITDATA variable assignment from the line above it. But it also suffers from the concern in my comment for the above line -- we want this to be the CSM_PATH value for the booted PIT node, not the current path.

install/bootstrap_livecd_usb.md
```diff
@@ -693,6 +693,8 @@ On first log in (over SSH or at local console), the LiveCD will prompt the admin
 
    ```bash
    pit# script -af /var/www/ephemeral/prep/admin/booted-csm-livecd.$(date +%Y-%m-%d).txt
+   pit# PITDATA=$(lsblk -o MOUNTPOINT -nr /dev/disk/by-label/PITDATA)
```
mharding-hpe, line 696: Why are we setting this variable here if it was already set in /etc/environment?
And the assignment here will only persist for this shell session -- any future shell sessions (as well as any subprocesses of this process) will use the value from the environment file.

install/bootstrap_livecd_usb.md
```diff
@@ -693,6 +693,8 @@ On first log in (over SSH or at local console), the LiveCD will prompt the admin
...
+   pit# CSM_PATH=${PITDATA}/${CSM_RELEASE}
```
mharding-hpe, line 697: Same concerns as on the preceding line.

mharding-hpe (changes requested):

rustydb:
The release/1.2 counterpart merged, but this PR still has conversations to resolve. A revert for #1433 is made at #1443.

install/bootstrap_livecd_remote_iso.md
rustydb: I see.

I'll fix this up by setting it in the above step where the other 2 variables are set.

install/bootstrap_livecd_usb.md
rustydb: I'll rework this to add these into the environment file later.
I'm doing this work in-tandem with a larger overhaul of these files (MTL-1695), and I want to eliminate the duplication of `/var/www/ephemeral`. The docs shouldn't know what that path is perse, that path can be resolved after the system has booted.

install/bootstrap_livecd_usb.md
rustydb: I actually didn't notice the usage of `/etc/environment` here on Friday, thank you for pointing that out.

install/bootstrap_livecd_usb.md
rustydb: Nice catch, thank you.

I've corrected this in my recent push.

johren-hpe (changes requested):

install/bootstrap_livecd_usb.md
```diff
@@ -569,7 +563,7 @@ Now that the configuration is generated, the LiveCD must be populated with the g
 1. Unmount the data partition:
```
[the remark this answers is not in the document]
johren-hpe: I don't think it would have.   They are and were set in the typescript session.   As soon as we type "exit" they go away.   I think it was the same issue despite all of the other changes.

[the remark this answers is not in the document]
johren-hpe: PITDATA would have been defined again (from /etc/environment) once the PIT was rebooted.  But these two commands (the cp and the umount) would not have it set since they are before the boot.

johren-hpe (approved):

mharding-hpe (approved):

PR #1573, opened 2022-05-09
CASMINST-4596: Remove some entries from the linter spelling file

mharding-hpe:
Backport of: https://github.com/Cray-HPE/docs-csm/pull/1572


.spelling
```diff
@@ -97,577 +55,120 @@
...
-c3
-ca.crt
-cabinets-yaml
-cabinets.yaml
-cable_info_disconnected
 cadvisor
-can-cidr
 CAN-to-CMN
-CAN.json
 Canu
```
johren-hpe, line 94: Should Canu always be CANU?

.spelling
```diff
@@ -97,577 +55,120 @@
...
-cfs-state-reporter
-changeme
-Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node
 channeled
-ChassisBMC
-ChassisBMCs
-ChassisSensors
 checkmark
-ChildVerificationFailed
-chn
```
johren-hpe, line 309: Should we have all of the capitalized network names in here?  NMN, CHN, CAN, CMN, HMN, NMNLB, HMNLB, NMN_RVR, HMN_RVR, NMN_MTN, HMN_MTN since these are acronyms rather than filenames, service names, etc.

johren-hpe (changes requested):

.spelling
mharding-hpe: Yeah, good call.
```suggestion
CANU
```

.spelling
mharding-hpe: Yeah, I removed it since it was lowercase, but you're right that it would make sense to instead convert those to all caps. I'll make that update.

johren-hpe (approved):

PR #1692, opened 2022-05-25
CASMNET-1543 Added xname check on HSN reservation before adding to CHN

shunr-hpe:


rustydb:
This was merged via #1709

PR #1779, opened 2022-06-08
Remove duplicated PKI block

pawsey-kbuckley:
Just noticed that the Index of the operations section,

https://cray-hpe.github.io/docs-csm/en-10/operations/

has the PKI topic list duplicated.

Are the indexes automatically generated?

Anyroad, this commit removes the duplication.


mharding-hpe (approved):

mharding-hpe:
> Are the indexes automatically generated?

They may have been originally, but they're also manually updated. Unless we start seeing this happen as a pattern, I'm not going to investigate to see when and how this duplication got introduced, although it does sound like a thrilling mystery. :)

PR #1813, opened 2022-06-10
CASMINST-4790: Reorganize and clarify NCN image modification instructions

mharding-hpe:
# Description

CASMTRIAGE-3411 was opened during the csm-1.2 install of tyr. The short version is that the NCN image modification script behaved in a way that the installer did not expect, and it resulted in a mismatch of SSH keys on the NCNs. This was not a bug in the script, but resulted because of how it was documented. After I determined the root cause of what had happened during that install, I opened CASMINST-4790 to address a couple of specific issues that helped cause the situation. However, in the course of making those changes I realized that the page as a whole is not easy to follow for people who are not that familiar with the process. So I used this PR to not only correct the specific problems we saw on tyr, but also to reorganize the page to make it more user friendly.

The actual procedure being following isn't significantly altered due to my changes. The biggest changes are:
- Added a pointer to the non-PIT version of this procedure, in case folks wanting to do it stumble across the wrong page.
- I removed the step to create the squashfs links. This step already exists on the deploy management nodes page, and is performed directly after this image modification step. So it was redundant.
- I slightly modified the examples given, mainly to stop using the find command (one of the causes of the tyr issue).
- I provided steps for the user to create their own root password hash, instead of just having the choice of using the PIT root password or entering the password during the script execution. Other than the small shell script wrapper (which just makes sure the user doesn't make a typo in the password), the password hashing command itself was lifted from a different operational procedure (so it's not untested code).
- I gave more details about how to view valid timezone options.
- I made it clear that both images should be modified with a single command to ensure they have matching SSH keys. Currently that is only kind of implied, in that our examples always both images.
- I removed the usage statement from the docs. Having it there meant that if we ever changed the script usage, we'd need to remember to update it here. I figured it was cleaner and more concise to just explain the high points, and to tell them how to run the script to view the usage.

I know we don't like to make changes like this so close to release, but I think this is one we need to make. I think that without this change, there will be some customer install issues. They are easy ones to solve, but they will cost time and cause frustration.

# Checklist Before Merging

### DIFF
--- a/operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
+++ b/operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
@@ -102,7 +102,7 @@ else
     read -s PW2
     echo
     if [[ $PW1 != $PW2 ]]; then
-        echo "ERROR: Passwords do not match"        
+        echo "ERROR: Passwords do not match"
     else
         export SQUASHFS_ROOT_PW_HASH=$(echo "$PW1" | openssl passwd -6 -salt $(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c4) --stdin)
         [[ -n $SQUASHFS_ROOT_PW_HASH ]] && echo "Password hash set and exported" || echo "ERROR: Problem generating hash"
--- a/operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
+++ b/operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
@@ -1,122 +1,171 @@
 # Set NCN Image Root Password, SSH Keys, and Timezone on PIT Node
 
-Modify the NCN images by setting the root password and adding SSH keys for the root account.
-Optionally, changing the timezone for the NCNs can also be done at this time. This procedure shows this process being
-done on the PIT node during a first time installation of the CSM software.
+> **Note:** This procedure is **required** during initial CSM installs **before management nodes are first deployed**.
 
-***Note:*** This procedure **is required** to be done during an initial CSM software installation
-**before management nodes are first deployed**.
+Modify the NCN images by setting the `root` user password and adding SSH keys for the `root` user account.
+If desired, also change the timezone for the NCNs.
 
-## Set the root password and add SSH keys to the NCN images
+This page describes this procedure being performed on the PIT node during a first time installation of the CSM software.
+If wanting to perform this operation after the initial CSM install, then
+see [Set NCN Image Root Password, SSH Keys, and Timezone](Change_NCN_Image_Root_Password_and_SSH_Keys.md).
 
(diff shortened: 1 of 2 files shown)


operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
```diff
@@ -1,122 +1,197 @@
...
-          environment variable (see below) along with either -d or -N.
+To have the script generate the SSH keys automatically, it must be provided with the `ssh-keygen` options to use.
 
-          -a             Do *not* modify the authorized_keys file in the squashfs.
-                         If modifying a previously modified image, or an
-                         authorized_keys file that contains the public key is already
-                         included in the directory used with the -d option, you may
-                         want to use this option.
+- To view the complete list of supported `ssh-keygen` options, view the script usage statement by running it with the `-h` argument.
+- If the `-N` option is not used to specify the passphrase, then the script will prompt for the passphrase when it generates the keys.
```
heemstra, line 81: That is true if you don't provide an argument to `-N`... but you may, optionally, provide that argument.  E.g. `-N ''` to have it be blank and not prompt.

heemstra (approved):

operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
mharding-hpe: Yeah, I cover that in one of the examples. Or you did, or whoever wrote the original doc -- I copied that part of that example. But anyway, I can add a sentence here too explaining that any passphrase (including a blank one) is a way to avoid the prompts.

mharding-hpe: Done

heemstra (approved):

operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
```diff
@@ -1,122 +1,199 @@
...
-                                              -s $(find . -name "storage-ceph-*.squashfs" | sort -V | tail -1)
-   ```
+> The script uses `read -s` to prevent the password from being echoed to the screen or saved
+> in the shell history. It unsets the plaintext password variables at the end, so that only
+> the hash is preserved.
 
-   The script will save the original SquashFS images in `./{k8s,ceph}/old`. The new image filenames will
-   have a `secure-` prefix. The `initrd` and kernel will retain their original filenames.
+```bash
+pit# \
```
rustydb, line 97: Can this prompt (and any others) move out of the snippet?

rustydb (approved):

operations/security_and_authentication/Change_NCN_Image_Root_Password_and_SSH_Keys_on_PIT_Node.md
mharding-hpe: Not for 1.2, I don't think, but for the main backport they will.

heemstra (approved):

PR #1859, opened 2022-06-16
Casmtriage 3539 12 internal ssh access execution failed

atifsyedali:
# Description

Fix failing SSH BICAN tests on Fanta:

- Replace `ssh` with `slogin` when using Mellanox switches.
- Properly recognize the command prompt for Mellanox switches
- When using hmnlb and nmnlb, previously all nodes to connect to were being normalized to hmcollector.* and api.* instead of their actual hostnames, which caused an issue with switches -- switch passwords did not work for NCNs. Now, hmcollector.* and api.* is only used when testing connection to NCNs. Negative tests for connecting to other node types via hmnlb and nmnlb still use the original hostnames of those node types.
- When connecting from a switch to another node, use "vrf Customer" only for CAN, CMN, and CHN but not for any other network.
- Skip tests when connecting to anything over `nmn` with Mellanox switches (see CASMNET-787)
- Use IPs for anything connecting from switches over CMN with Aruba switches (see CASMNET-1599)
- Previously, was not counting the test for which node cannot be found, but did display a FAILED message for it. That was confusing ("there are X failed tests but summary only shows Y", where X > Y). Changed it so that the test with no nodes found is still counted.
- Removed UAN and CN as defaults in node types to test from and to. These may not be ready by the time we validate CSM health.
- Some improvements to code quality and structure.

# Checklist Before Merging

- [ ] My commits or Pull-Request Title contain my JIRA information, or I don't have a JIRA.

[1]: https://github.com/Cray-HPE/docs-csm/blob/MTL-1695/introduction/documentation_conventions.md#using-prompts


scripts/operations/pyscripts/pyscripts/core/ssh/ssh_host.py
```diff
@@ -76,53 +77,85 @@ def get_password(self, reset_cached = False):
...
 
-        (2) is a workaround that substitutes the IP of the target host instead of the hostname to connect.
-        It is so we can get around the bug https://jira-pro.its.hpecorp.net:8443/browse/CASMNET-1600
-
-        Once that bug is resolved, either this entire method and the logic to use this method in ssh_connection should be removed,
-        or everything except the last line in this method that returns target_ssh_host.hostname should be removed.
+        2. If using nmnlb or hmnlb, then always goes to api.nmnlb.<system-domain> and api.hmnlb.<system-domain> respectively.
         """
-        if self.vrf and "cmn." in target_ssh_host.domain_suffix and "switch" in target_ssh_host.type:
+        if self.is_aruba_switch() and "cmn." in target_ssh_host.domain_suffix:
```
johren-hpe, line 154: I think this is okay for now but I think we could hit a case where the domain suffix is something like `can.eniacmn.com`.   That would falsely match the "cmn." case.

atifsyedali: Yes totally agree with you. At the very least I can split by period and compare. Will make the change.

rustydb (approved):

PR #1861, opened 2022-06-16
CASMINST-4821: Validate SLS health before using CSI to set no-wipe during node rebuilds

mharding-hpe:
# Description

[CASMTRIAGE-3255](https://jira-pro.its.hpecorp.net:8443/browse/CASMTRIAGE-3255) documents a case we have seen twice now where SLS is not healthy when calling CSI to set the no-wipe flag, resulting in the upgrade hanging. Currently the root cause of the issue is not understood.

In order to provide a better failure condition (rather than an infinite loop), this PR adds a very simple SLS health check before making the CSI call to set the no-wipe flag.

In order to better isolate the problem, this PR also adds this same check just before the "drain node" step (because it would be helpful to know if it is the drain which is causing this problem).

At Di's suggestion, I added this logic for node rebuilds both inside and outside of the upgrade context.

The SLS health check consists of 2 API calls, to the liveness and health endpoints. It validates that the correct response status is given (204 and 200 respectively). To avoid cases where SLS will recover on its own, it will retry the SLS health checks for up to 5 minutes before declaring it a failure.

I have tested the SLS health check function on hela (csm-1.2) and frigg (csm-1.0) and  validated that it works as expected. I haven't tested it in a live upgrade.

# Checklist Before Merging

### DIFF
--- a/upgrade/1.2/scripts/common/ncn-common.sh
+++ b/upgrade/1.2/scripts/common/ncn-common.sh
@@ -81,6 +81,43 @@ export TARGET_MGMT_XNAME=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https
 
 export TARGET_IP_NMN=$(dig +short $TARGET_NCN.nmn)
 
+# Just do basic API calls to SLS to make sure that it is responding.
+function check_sls_health() {
+    local timeout first
+
+    echo "Checking SLS health..."
+
+    # To make sure that this is not a case where it will get better by itself,
+    # we will retry for up to 5 minutes before quitting.
+    let timeout=SECONDS+300
+    first=Y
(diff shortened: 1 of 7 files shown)


kimjensen-hpe (approved):

dzou-hpe:
`drain_node` is also used for rebuild automation too, so maybe we should add this new check to inside the `drain_node` function?
or we should add it in  "upgrade/1.2/scripts/rebuild/ncn-rebuild-master-nodes.sh" and  "upgrade/1.2/scripts/rebuild/ncn-rebuild-worker-nodes.sh" too

dzou-hpe (approved):

dzou-hpe (approved):

bklei (approved):

PR #1866, opened 2022-06-17
CASMINST-4820: Add troubleshooting/remediation steps for SLS postgres SyncFailed

mharding-hpe:
# Description

[CASMTRIAGE-3255](https://jira-pro.its.hpecorp.net:8443/browse/CASMTRIAGE-3255) documents a case we have seen twice now where SLS is not healthy when calling CSI to set the no-wipe flag, resulting in the upgrade hanging. Currently the root cause of the issue is not understood.

[CASMINST-4821]https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-4821) is adding better checking for this condition to the upgrade scripts.

This PR is adding documentation on how to detect if a system is in this situation, and the remediation steps that have worked in the cases where this was seen.

I wrote up the procedure so that it should apply whether this is seen during an upgrade or otherwise. Until we know the root cause, we don't know if this can only happen during upgrades, or if it can happen during any node rebuild. If we do get clarity on this, the page can be refined accordingly.

# Checklist Before Merging

### DIFF
--- a/troubleshooting/index.md
+++ b/troubleshooting/index.md
@@ -9,6 +9,7 @@ This document provides links to troubleshooting information for services and fun
   * [`initrd.img.xz` not found](known_issues/initrd.img.zx_not_found.md)
 * [Platform CA Issues](known_issues/platform_ca_issues.md)
 * [Kafka Failure after CSM 1.2 Upgrade](known_issues/kafka_upgrade_failure.md)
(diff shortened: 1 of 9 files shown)


kimjensen-hpe (approved):
looks great

bklei (approved):

PR #1869, opened 2022-06-17
CASMTRIAGE-3536: time hack

heemstra:
# Description

This commit makes the following changes:

1. In upgrade/prerequisites.sh, correct/verify the chrony config
for every node. Do this instead of doing it one node at a time
as they are upgraded. After `csm_ntp.py` has been run on a node,
use `chronyc makestep` to force time to sync immediately. With
this approach, time should be in sync across the cluster prior
to the upgrade beginning in earnest.

2. Once an upgraded node is online and settled, use `chronyc makestep`
to force time to sync immediately.

### DIFF
--- a/upgrade/1.2/scripts/common/ncn-rebuild-common.sh
+++ b/upgrade/1.2/scripts/common/ncn-rebuild-common.sh
@@ -49,34 +49,24 @@ else
 fi
 
 state_name="ELIMINATE_NTP_CLOCK_SKEW"
-state_recorded=$(is_state_recorded "${state_name}" "${target_ncn}")
-if [[ $state_recorded == "0" && $2 != "--rebuild" ]]; then
-    echo "====> ${state_name} ..."
+state_recorded=$(is_state_recorded "${state_name}" "$target_ncn")
+in_sync=$(ssh "${target_ncn}" timedatectl | awk /synchronized:/'{print $NF}')
+if [[ "$in_sync" == "no" ]]; then
     {
-    if [[ "${target_ncn}" != "ncn-m001" ]]; then
-      # ensure the directory exists
-      ssh "${target_ncn}" 'mkdir -p /srv/cray/scripts/common/'
-      # copy the NTP script and template to the node
-      scp -pr "${CSM_ARTI_DIR}"/chrony/ "${target_ncn}":/srv/cray/scripts/common/
-      # run the script
-      #shellcheck disable=SC2029
-      if ! ssh "${target_ncn}" "TOKEN=$TOKEN /srv/cray/scripts/common/chrony/csm_ntp.py"; then
-          echo "${target_ncn} csm_ntp failed"
-          exit 1
-      fi
-
-      # if the node is not in sync after two minutes, fail
-      if ! ssh "${target_ncn}" 'chronyc waitsync 6 0.5 0.5 20'; then
-          echo "${target_ncn} the clock is not in sync.  Wait a bit more or try again."
-          exit 1
-      else
-          record_state "${state_name}" "${target_ncn}"
-      fi
+    ssh "$target_ncn" chronyc makestep
+    sleep 5
+    in_sync=$(ssh "${target_ncn}" timedatectl | awk /synchronized:/'{print $NF}')
+    if [[ "$in_sync" == "no" ]]; then
+        echo "The clock on ${target_ncn} is not in sync.  Wait a bit more or try again."
+        exit 1
+    else
+        record_state "${state_name}" "${target_ncn}"
     fi
     } >> ${LOG_FILE} 2>&1
 else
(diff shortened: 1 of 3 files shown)


heemstra:
Testing in progress

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
```diff
@@ -308,6 +278,9 @@ if [[ $target_ncn != ncn-s* ]]; then
     } >> ${LOG_FILE} 2>&1
 fi 
 
+# force time to sync immediately
+ssh "${target_ncn}" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null chronyc makestep
```
mharding-hpe, line 282: Will that fail if it is not able to synchronize the times? We wouldn't want to proceed if the sync fails.

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
```diff
@@ -48,36 +48,6 @@ else
     echo "====> ${state_name} has been completed"
 fi
 
-state_name="ELIMINATE_NTP_CLOCK_SKEW"
```
mharding-hpe, line 51: Would we not want to also check the sync here? I know we check it at the prereq step, but in some upgrades there may be days between when the prereq script is run and any given node is rebuilt. It seems long enough that we'd want to verify the clocks have not drifted, right?

heemstra: Yeah, I can do the `makestep` here and also check.  We shouldn't have to re-run csm_ntp.py though.

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
heemstra: I can add a check here.  This command by itself won't do it.

rustydb (approved):

dzou-hpe (approved):

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
```diff
@@ -308,6 +298,16 @@ if [[ $target_ncn != ncn-s* ]]; then
     } >> ${LOG_FILE} 2>&1
 fi 
 
+# force time to sync immediately
+{
+  ssh "${target_ncn}" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null chronyc makestep
+  sleep 1
+  in_sync=$(ssh "${target_ncn}" timedatectl | awk /synchronized:/'{print $NF}')
+  if [[ "$in_sync" = "no" ]]; then
```
bklei, line 306: double equal?

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
```diff
@@ -357,3 +357,9 @@ if [[ ${target_ncn} != ncn-s* ]]; then
         echo "====> ${state_name} has been completed"
     fi
 fi
+
+if [[ "$in_sync" = "no" ]]; then
```
bklei, line 361: double equal?

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
```diff
@@ -49,35 +49,25 @@ else
...
-      if ! ssh "${target_ncn}" 'chronyc waitsync 6 0.5 0.5 20'; then
-          echo "${target_ncn} the clock is not in sync.  Wait a bit more or try again."
-          exit 1
-      else
-          record_state "${state_name}" "${target_ncn}"
-      fi
+    ssh "$target_ncn" chronyc makestep
+    sleep 1
+    in_sync=$(ssh "${target_ncn}" timedatectl | awk /synchronized:/'{print $NF}')
+    if [[ "$in_sync" = "no" ]]; then
```
bklei, line 59: double equal?

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
heemstra: I read up on that.  `==` is a bash-ism for `=`.  I'm being inconsistent here though, so I'll fix.

upgrade/1.2/scripts/common/ncn-rebuild-common.sh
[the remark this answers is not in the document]
heemstra: fixing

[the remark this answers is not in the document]
heemstra: fixed

denniswalker:
Merging given clock skew issues are occurring everywhere and we can easily revert this PR if there's an issue.

PR #1959, opened 2022-06-30
CASMINST-4883: Provide details on restoring CFS configuration for NCN worker nodes

deantroe:
# Description

Describe how the upgrade procedure clears the CFS configuration from worker nodes
and how to restore required configuration for other products to ensure the compute
nodes and UANs remain functional during the upgrade.

# Checklist Before Merging

### DIFF
--- a/upgrade/1.2/Stage_0_Prerequisites.md
+++ b/upgrade/1.2/Stage_0_Prerequisites.md
@@ -282,7 +282,7 @@ If the following command does not complete successfully, check if the `TOKEN` en
    If the failure persists, then open a support ticket for guidance before proceeding.
 
    `prerequisites.sh` clears the existing CFS configuration for each Management node. As each
-   Management node is upgraded, documentation will refer to the CFS configuration that should be
+   worker node is upgraded, the documentation will refer to the CFS configuration that should be
    assigned to the node at that time. If any Management node is unexpectedly rebooted prior to this,
    CFS will not automatically personalize the node after it has booted.
 
--- a/upgrade/1.2/Stage_0_Prerequisites.md
+++ b/upgrade/1.2/Stage_0_Prerequisites.md
@@ -281,6 +281,11 @@ If the following command does not complete successfully, check if the `TOKEN` en
    [Upgrade Troubleshooting](README.md#relevant-troubleshooting-links-for-upgrade-related-issues).
    If the failure persists, then open a support ticket for guidance before proceeding.
 
+   `prerequisites.sh` clears the existing CFS configuration for each Management node. As each
+   Management node is upgraded, documentation will refer to the CFS configuration that should be
+   assigned to the node at that time. If any Management node is unexpectedly rebooted prior to this,
+   CFS will not automatically personalize the node after it has booted.
+
 1. Unset the `NEXUS_PASSWORD` variable, if it was set in the earlier step.
 
    ```bash
--- a/upgrade/1.2/Stage_0_Prerequisites.md
+++ b/upgrade/1.2/Stage_0_Prerequisites.md
@@ -283,7 +283,8 @@ If the following command does not complete successfully, check if the `TOKEN` en
 
    `prerequisites.sh` clears the existing CFS configuration for each Management node. As each
    worker node is upgraded, the documentation will refer to the CFS configuration that should be
-   assigned to the node at that time. If any Management node is unexpectedly rebooted prior to this,
+   assigned to the node at that time. If any worker node is unexpectedly rebooted prior to this, or if any other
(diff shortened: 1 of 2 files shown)


jsl-hpe (approved):

upgrade/1.2/Stage_0_Prerequisites.md
```diff
@@ -270,6 +270,11 @@ If the following command does not complete successfully, check if the `TOKEN` en
    [Upgrade Troubleshooting](README.md#relevant-troubleshooting-links-for-upgrade-related-issues).
    If the failure persists, then open a support ticket for guidance before proceeding.
 
+   `prerequisites.sh` clears the existing CFS configuration for each Management node. As each
+   Management node is upgraded, documentation will refer to the CFS configuration that should be
```
mharding-hpe, line 274: I don't think this is accurate. It is true that CFS is cleared for all Management nodes, but our docs do not say anything about CFS as the nodes are upgraded. We handle CFS as the final step of the upgrade process.

I see that this PR is adding this in stage 2 for the worker nodes, but not for the master nodes (also in stage 2) or storage nodes (in stage 1). Should we also be adding callouts for CFS for those nodes as they are upgraded, or should we revise this paragraph to be clear that we only call out the CFS stuff as the worker nodes are upgraded.

upgrade/1.2/Stage_2.md
```diff
@@ -59,6 +59,16 @@
 
    > **NOTE:** The root password for the node may need to be reset after it is rebooted.
 
+1. Assign a new CFS configuration to the worker node.
+
+   The content of the new CFS configuration is described in _HPE Cray EX System Software Getting Started Guide S-8000_, section
+   "HPE Cray EX Software Upgrade Workflow" subsection "Cray System Management (CSM)". Replace `${NEW_NCN_CONFIGURATION}` with
+   the name of the new CFS configuration and `${XNAME}` with the xname of ncn-w001.
```
mharding-hpe, line 66:
```suggestion
   the name of the new CFS configuration and `${XNAME}` with the component name (xname) of the worker node that was upgraded.
```

mharding-hpe (changes requested):

upgrade/1.2/Stage_0_Prerequisites.md
deantroe: The "documentation will refer to the CFS configuration" part was a reference to the other blurb I am adding with this PR, specifically the callout in Stage 2 to the GSG which tells them how to put the config back in place.  And then, like you said, existing Stage 5 (IIRC) documentation goes into detail on the CSM configuration.

I am unaware of any critical CFS configuration that needs to be applied to master and storage nodes in this "gap" that exists between the time when CSM upgrades wipe configuration and later in the overall upgrade process when each product adds their configuration layer back in.  The point of this doc update is to cover this gap period so the HSN and DVS on worker nodes continue to function.  If someone is aware of CFS layers that are critical for master and/or storage nodes, I am happy to expand the documentation.  Internal testing, AFAIK, was done with the worker change only.

mharding-hpe: So in that case I'd change the wording of this paragraph. Something like:
```suggestion
   worker node is upgraded, the documentation will refer to the CFS configuration that should be
```

upgrade/1.2/Stage_0_Prerequisites.md
```diff
@@ -270,6 +270,11 @@ If the following command does not complete successfully, check if the `TOKEN` en
...
+   assigned to the node at that time. If any Management node is unexpectedly rebooted prior to this,
```
mharding-hpe, line 275:
```suggestion
   assigned to the node at that time. If any Management node is unexpectedly rebooted prior to the end of the CSM upgrade,
```
```suggestion
   assigned to the node at that time. If any Management node is unexpectedly rebooted prior to this,
```

deantroe: The timeframe when a node would be rebooted and not run CFS node personalization is after ```prerequisites.sh``` runs and
- worker nodes: before the new instructions in Stage 2 tell the admin how to put the initial/critical CFS configuration back in place
- storage and master nodes: until the end of the CSM update

Given that, I am not sure the suggestion is better than the original text.  Thoughts?

rustydb: "this" does imply a more immediate context

rustydb (approved):

upgrade/1.2/Stage_0_Prerequisites.md
mharding-hpe: How about my updated wording?

PR #2014, opened 2022-07-08
CASMINST-4924: move argo templates into docs

dzou-hpe:
# Description

move argo templates into docs

https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-4924

# Checklist Before Merging


alexanderkingh (approved):

heemstra:
It'd be better to have the bash in individual script files so they can be linted - and then have the yaml invoke the scripts.

dzou-hpe:
> It'd be better to have the bash in individual script files so they can be linted - and then have the yaml invoke the scripts.

Agreed, the middle term plan is to reduce bash to minimum and long term plan is to get rid of bash here. We are also trying to avoid ssh as much as possible so we don't want to put it on disk. I will need to develop a new way for storing bash outside then `import` it on the fly.

workflows/ncn/worker/common.envar.yaml
```diff
@@ -0,0 +1,44 @@
...
+TARGET_NCN={{ `{{inputs.parameters.targetNcn}}` }}
+TARGET_XNAME=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/apis/sls/v1/search/hardware?extra_properties.Role=Management" | \
+     jq -r ".[] | select(.ExtraProperties.Aliases[] | contains(\"$TARGET_NCN\")) | .Xname")
+TARGET_MGMT_XNAME=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/apis/sls/v1/search/hardware?extra_properties.Role=Management" | \
+  jq -r ".[] | select(.ExtraProperties.Aliases[] | contains(\"$TARGET_NCN\")) | .Parent")
+
+TARGET_NCN_mgmt_host="${TARGET_NCN}-mgmt"
+
+
+export IPMI_USERNAME=$(curl -XGET -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/apis/scsd/v1/bmc/creds"| jq -r ".Targets[] | select(.Xname | contains(\"$TARGET_MGMT_XNAME\")) | .Username")
```
mharding-hpe, line 41: I don't know if the export in this argo template is the same as a normal Bash export, but if so, I don't think we need to export the username. Only the password needs to be exported.

mharding-hpe, line 41: I don't know if the export in this argo template is the same as a normal Bash export, but if so, I don't think we need to export the username. Only the password needs to be exported.

```suggestion
IPMI_USERNAME=$(curl -XGET -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/apis/scsd/v1/bmc/creds"| jq -r ".Targets[] | select(.Xname | contains(\"$TARGET_MGMT_XNAME\")) | .Username")
```

mharding-hpe:
> > It'd be better to have the bash in individual script files so they can be linted - and then have the yaml invoke the scripts.
> 
> Agreed, the middle term plan is to reduce bash to minimum and long term plan is to get rid of bash here. We are also trying to avoid ssh as much as possible so we don't want to put it on disk. I will need to develop a new way for storing bash outside then `import` it on the fly.

But as long as we still have Bash, would it not be possible to do as John suggests?

dzou-hpe:
> > > It'd be better to have the bash in individual script files so they can be linted - and then have the yaml invoke the scripts.
> > 
> > 
> > Agreed, the middle term plan is to reduce bash to minimum and long term plan is to get rid of bash here. We are also trying to avoid ssh as much as possible so we don't want to put it on disk. I will need to develop a new way for storing bash outside then `import` it on the fly.
> 
> But as long as we still have Bash, would it not be possible to do as John suggests?

not at the moment and we have ticket open for revisit all the bash we use here but we never have time to do it. I'd suggest bringing this up in standup, if we can have more time on this, i'd love to work on that.

workflows/ncn/worker/worker.rebuild.yaml
```diff
@@ -0,0 +1,246 @@
...
+              parameters:
+              - name: targetNcn
+                value: {{$value}}
+              - name: dryRun
+                value: "{{$.DryRun}}"
+          # wipe and reboot: parallel
+          #     once a worker node is drained from k8s
+          #     we can safely wipe and reboot this node
+          #     regardless of what state other nodes are
+          - name: wipe-and-reboot-{{$value}}
```
rustydb, line 188: We don't have to wipe anymore, we can though.

This should probably just be "set nextboot"

It should just set pxe for the next boot, and maybe disable the disk bootloader.

For reference the rebuild script in main is updated to do just this, I'd link it for reference right now if I could but I'm mobile. I'll link it Monday unless someone beats me to it.

rustydb (changes requested):

workflows/ncn/worker/worker.rebuild.yaml
rustydb: Wiping would ensure a pxe boot, but the reason I care is I want to remove the mentality that a wipe is required. The deployment code will wipe. If a wipe fails then it's the fault of the deployment code, and the deployment code handles wiping better than the manual/argo step(s).

dzou-hpe: i still have that ticket assigned to me. the reason why i kept this for now is so i can test on a 1.2 system until we get all 1.3 upgrade script cleaned up

rustydb (approved):

PR #2127, opened 2022-07-26
CASMPET-5784: Add Nexus export to Prerequisites.sh

ndavidson-hpe:
# Description

This adds the Nexus export to prerequisites.sh. A concern with this is the export will add significant time to the running time of the script. It takes around one hour for every 60 Gigabytes in Nexus.

# Checklist Before Merging

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -298,6 +298,21 @@ else
     echo "====> ${state_name} has been completed"
 fi
 
+state_name="EXPORT_NEXUS"
+#shellcheck disable=SC2046
+state_recorded=$(is_state_recorded "${state_name}" $(hostname))
+if [[ $state_recorded == "0" && $(hostname) == "ncn-m001" ]]; then
+    echo "====> ${state_name} ..."
+    {
+    ${locOfScript}/../../../scripts/nexus-export.sh
+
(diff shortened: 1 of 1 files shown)


bklei (approved):

kimjensen-hpe (commented):
Given your comments in the Description - should there be some wording on how long this ie expected to take. Can we get the size of nexus and determine the approx length of time for the export and add that as output that the user would see?

rgross727 (approved):

ndavidson-hpe:
> Given your comments in the Description - should there be some wording on how long this ie expected to take? Can we get the size of nexus and determine the approx length of time for the export and add that as output that the user would see?

I don't think there is a way to give a good length. With a lot of testing on different hardware we could get an estimate but its using tar compression so depending on hardware and overall system availability it could change. My one minute per Gigabyte is an estimate based on one or two systems with both about 100 Gigabytes there so as Nexus grows Im not sure if it linearly scales.

dzou-hpe:
@djlapenta @atifsyedali  looks like we are adding `one hour for every 60 Gigabytes in Nexus.`. i think prereq is going to be a hot topic that we should let argo run it so we can run things in parallel.

ndavidson-hpe:
I think this should be held back for now to discuss more.

ndavidson-hpe:
This is not being added directly to prereq.sh

PR #2159, opened 2022-08-02
CASMINST-5064 NCN 1.1.1 Validation Script

mantoanibroadwing:
# Description

Implemented script for NCN 1.1.1 validation fully automated


scripts/operations/validation/ncn111.sh
```diff
@@ -0,0 +1,114 @@
...
+
+check_spire_agent() {
+    echo "Checking spire agent status"
+    spire=$(kubectl get pods -n spire | grep request-ncn-join-token)
+    if [[ $? = 0 ]]; then
+        echo "No issues found."
+        echo "---------------------------------------"
+        return 0
+    else
+        echo "Issues found"
```
atifsyedali, line 83: I think it would be really helpful if you pipe the commands output to a file, then either cat the file here or reference it in the message above, instead of simply saying "Issues found".

This would mean that, for remediation, the user does not have to re-run this command to figure out what actually is wrong. The remediation doc becomes much smaller then, and we can even skip remediation automation as it is lower priority then validation (because validation runs 100% of the time, remediation runs <100% of the time).

This comment applies to all future validation scripts.

scripts/operations/validation/ncn111.sh
```diff
@@ -0,0 +1,114 @@
...
+
+cfs_state_reporter() {
+    echo "Checking cfs-state-reporter on HTTPS connection pool"
+    cfs_check=$(systemctl status cfs-state-reporter | grep HTTPSConnectionPool)
+    if [[ $? != 0 ]]; then
+        echo "No issues found."
+        echo "---------------------------------------"
+        return 0
+    else
+        echo "Issues found"
```
atifsyedali, line 96: Please see my comment above regarding piping to output file.

scripts/operations/validation/ncn111.sh
```diff
@@ -0,0 +1,114 @@
...
+
+clock_skew() {
+    echo "Checking clock skew"
+    clocskew_check=$(chronyc sources -v)
+    if [[ $? = 0 ]]; then
+        echo "No issues found."
+        echo "---------------------------------------"
+        return 0
+    else
+        echo "Issues found, proceed to https://github.com/Cray-HPE/docs-csm/blob/CASMINST-5064/troubleshooting/known_issues/issues_with_ncn_health_checks.md"
```
atifsyedali, line 109: Please see my two comments above regarding (1) not using PR branch's permalink in production code; (2) piping to output file.

troubleshooting/known_issues/issues_with_ncn_health_checks.md
```diff
@@ -0,0 +1,125 @@
+#### Known issues with NCN health checks
```
atifsyedali, line 1: As per my comments above, this file needs to be structured into using headings so that we can directly link to specific sections as per type of check failure.

atifsyedali (commented):
Added my review comments.

scripts/operations/validation/ncn111.sh
```diff
@@ -0,0 +1,114 @@
...
+    count_total=4
+    count_failed=$(($count_total - $count_success))
+
+    echo " "
+    echo "======================================="
+    echo " "
+    echo "Summary: $count_failed of $count_total tests FAILED"
+    if [[ "$velero_count" == 0 ]]; then
+        echo " "
+        echo "- Velero failed backups found, proceed to https://github.com/Cray-HPE/docs-csm/blob/CASMINST-5064/troubleshooting/known_issues/issues_with_ncn_health_checks.md for remediation steps"
```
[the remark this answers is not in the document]
denniswalker: I’m wondering about the solution for the branch name in the runbook URLs. Could we autodetermine which CSM release it is and inject a parameter into the URL?  Alternatively, the script could probably locate the local version of the runbook from itself.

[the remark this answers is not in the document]
atifsyedali: We have a variable for the release itself somewhere on the system, so we can link directly to the github.io page under that release.

The other option is that we reference the docs-flow in-cluster URL...

PR #2223, opened 2022-08-12
CASMPET-5889

dzou-hpe:
# Description

- CASMPET-5889  enable pipefail
- CASMPET-5889  add curl error handling
- CASMINST-5197 upload workflow files as part of nls deployment

# Testing
Fanta

# Checklist Before Merging

- [ ] My commits or Pull-Request Title contain my JIRA information, or I don't have a JIRA.

### DIFF
--- a/upgrade/scripts/upgrade/util/nls-keycloak-configure.py
+++ b/upgrade/scripts/upgrade/util/nls-keycloak-configure.py
@@ -22,6 +22,7 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 import argparse
+import os
 import sys
 import requests
 import json
@@ -34,24 +35,21 @@ def globalErrorHandler(error):
 urllib3.disable_warnings()
 
 parser = argparse.ArgumentParser(description='utility script to configure keycloak for cray-nls')
-parser.add_argument('--kc-client-id',required=True, help='keycloak admin client id')
-parser.add_argument('--kc-username',required=True, help='keycloak admin username')
-parser.add_argument('--kc-password',required=True, help='keycloak admin password')
-parser.add_argument('--argo-url',required=True, help='Public URL of argo')
-
-args = parser.parse_args()
 
 session = requests.Session()
 session.verify = False
 
-
+argo_url=os.environ.get('ARGO_URL')
+kc_client_id=os.environ.get('KC_CLIENT_ID')
+kc_username=os.environ.get('KC_USERNAME')
+kc_password=os.environ.get('KC_PASSWORD')
 try:
     # Get TOKEN
     data = {
             "grant_type":"password", 
-            "client_id": args.kc_client_id, 
-            "username": args.kc_username,
-            "password": args.kc_password
+            "client_id": kc_client_id, 
+            "username": kc_username,
+            "password": kc_password
             }
     token_response = session.post('https://api-gw-service-nmn.local/keycloak/realms/master/protocol/openid-connect/token', data=data)
     token_json = token_response.json()
@@ -75,12 +73,12 @@ try:
     
     needConfigure=False
     # Configure webOrigins
-    argoWebOrigin="https://"+args.argo_url
+    argoWebOrigin="https://"+argo_url
     if not argoWebOrigin in customerManagementClient['webOrigins']:
         customerManagementClient['webOrigins'].append(argoWebOrigin)
         needConfigure=True
     # Configure redirectUris
-    argoRedirectUri="https://"+args.argo_url+"/oauth/callback"
+    argoRedirectUri="https://"+argo_url+"/oauth/callback"
     if not argoRedirectUri in customerManagementClient['redirectUris']:
         customerManagementClient['redirectUris'].append(argoRedirectUri)
         needConfigure=True
--- a/upgrade/scripts/upgrade/util/nls-keycloak-configure.py
+++ b/upgrade/scripts/upgrade/util/nls-keycloak-configure.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2021-2022 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -22,7 +22,6 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 import argparse
-import http
 import sys
 import requests
 import json
@@ -45,19 +44,15 @@ args = parser.parse_args()
 session = requests.Session()
 session.verify = False
 
-# Get TOKEN
-# grant_type=password \
-#     --data-urlencode client_id="$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.client-id}' -n services | base64 -d)" \
-#     --data-urlencode username="$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.user}' -n services | base64 -d)" \
-#     --data-urlencode password="$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.password}' -n services | base64 -d)" \
-#     https://api-gw-service-nmn.local/keycloak/realms/master/protocol/openid-connect/token
-data = {
-        "grant_type":"password", 
-        "client_id": args.kc_client_id, 
-        "username": args.kc_username,
-        "password": args.kc_password
-        }
+
 try:
+    # Get TOKEN
+    data = {
+            "grant_type":"password", 
+            "client_id": args.kc_client_id, 
+            "username": args.kc_username,
+            "password": args.kc_password
+            }
     token_response = session.post('https://api-gw-service-nmn.local/keycloak/realms/master/protocol/openid-connect/token', data=data)
     token_json = token_response.json()
     token = token_json['access_token']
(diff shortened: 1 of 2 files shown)


kimjensen-hpe (approved):

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
```diff
@@ -50,38 +50,57 @@ function patchKeycloak() {
...
-    redirectUris=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients" | jq -r '.[] | select(.clientId=="oauth2-proxy-customer-management") | .redirectUris')
-    #shellcheck disable=SC2126
-    foundArgoInWebOrigins=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients" | jq '.[] | select(.clientId=="oauth2-proxy-customer-management") | .webOrigins' | grep "https://argo.cmn" | wc -l)
-    webOrigins=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients" | jq -r '.[] | select(.clientId=="oauth2-proxy-customer-management") | .webOrigins')
 
-    id=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients" | jq -r '.[] | select(.clientId=="oauth2-proxy-customer-management") | .id')
+    res_file=$(mktemp)
+    http_code=$(curl -s -o "${res_file}" -w "%{http_code}" -k -XGET -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients")
+    if [[ ${http_code} -ne 200 ]]; then
```
johren-hpe, line 56: We aren't going to hit this if the curl fails because of the set -e, right?   That will cause it to exit right after line 55.

johren-hpe (changes requested):

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
johren-hpe: Did you test any negative tests where it hits the error conditions?

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
```diff
@@ -50,38 +50,57 @@ function patchKeycloak() {
...
+    if jq '.[] | select(.clientId=="oauth2-proxy-customer-management") | .webOrigins'  < "${res_file}" | grep "https://argo.cmn"; then
+        echo "  Argo URL is set in WebOrigins"
+    else
+        tmpFile=$(mktemp)
+        jq ".webOrigins[.webOrigins | length] |= . + \"https://$argoUrl\"" < client.json  | tee "${tmpFile}"  > /dev/null
+        mv "${tmpFile}" client.json
     fi
 
-    curl -k -XPUT \
+    res_file=$(mktemp)
```
johren-hpe, line 92: You should clean up this tmp file.   I see 18000 temp files on fanta from the weekend testing.
```
ncn-m001:~ # ls /tmp/tmp.* | wc -l
18814
```

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
johren-hpe: It will if the curl succeeds and returns something like 404 or 500.  But it will not work, for example, if DNS is not working and it cannot resolve api-gw-service-nmn.local.

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
```diff
@@ -50,38 +50,57 @@ function patchKeycloak() {
...
+    echo "Get client id: ${id}"
 
-    curl -s -k -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients/${id}" > client.json
-    if [[ ${foundArgoInRedirectUris} -ge 1 ]];then
-        echo "Argo URL is set in RedirectUris"
-        echo "${redirectUris}"
+    http_code=$(curl -s -o "client.json" -w "%{http_code}" -k -XGET -H "Authorization: Bearer ${TOKEN}" "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients/${id}")
+    if [[ ${http_code} -ne 200 ]]; then
+        echo "Request Failed, Response code: ${http_code}"
+        cat "${res_file}"
```
johren-hpe, line 68: Is it res_file or client.json that you want to cat in this case?

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
johren-hpe: One other concern that I have is that it looks like prerequisites.sh will not fail when upgrade-cray-nls.sh fails because it is in a subshell.   I think it will just move on to the next stage and we won't notice that something went wrong.   Am I reading that wrong?

dzou-hpe: you are right ...... i might just throw out shell really... why can't it give me an error object like a real language would .... i need to babysit every tiny bits. I hate shell/bash more and more everyday

johren-hpe: I think if you just check ahead at the beginning to make sure the TOKEN is valid, that should cover the case where it cannot reach the gateway because of networking or DNS.

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
```diff
@@ -50,38 +50,57 @@ function patchKeycloak() {
     --data-urlencode username="$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.user}' -n services | base64 -d)" \
     --data-urlencode password="$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.password}' -n services | base64 -d)" \
     https://api-gw-service-nmn.local/keycloak/realms/master/protocol/openid-connect/token | jq -r '.access_token')
```
johren-hpe, line 52: Should also check here for a valid token.

johren-hpe (changes requested):

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
dzou-hpe: throw out bash, problem solved

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
dzou-hpe: throw out bash, problem solved

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
dzou-hpe: throw out bash, problem solved

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
dzou-hpe: throw out bash, problem solved

upgrade/scripts/upgrade/util/nls-keycloak-configure.py
```diff
@@ -0,0 +1,98 @@
...
+def globalErrorHandler(error):
+    print(e)
+    sys.exit("something went wrong")
+
+urllib3.disable_warnings()
+
+parser = argparse.ArgumentParser(description='utility script to configure keycloak for cray-nls')
+parser.add_argument('--kc-client-id',required=True, help='keycloak admin client id')
+parser.add_argument('--kc-username',required=True, help='keycloak admin username')
+parser.add_argument('--kc-password',required=True, help='keycloak admin password')
```
mharding-hpe, line 39: Would be better not to have this passed on the command line, but instead read from an environment variable.

upgrade/scripts/upgrade/util/upgrade-cray-nls.sh
```diff
@@ -44,44 +44,15 @@ function deployNLS() {
...
-    fi
-
-    curl -k -XPUT \
-        -H "Authorization: Bearer ${TOKEN}" \
-        -H "Content-Type: application/json" \
-        -d @client.json \
-        "https://api-gw-service-nmn.local/keycloak/admin/realms/shasta/clients/${id}" > /dev/null
+    kcClientId=$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.client-id}' -n services | base64 -d)
+    kcUsername=$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.user}' -n services | base64 -d)
+    kcPassword=$(kubectl get secrets keycloak-master-admin-auth -o jsonpath='{.data.password}' -n services | base64 -d)
```
mharding-hpe, line 49: This should be exported and read in as an environment variable, rather than passing it on the command line.

mharding-hpe (changes requested):
Does this need a 1.3 PR as well?

upgrade/scripts/upgrade/util/nls-keycloak-configure.py
dzou-hpe: i don't think that's necessary. we ssh as root and only root can get this credentials anyways because you won't have access to kubeconfig file

dzou-hpe: i am assuming we are trying to protect plain text password here? if that is the case, even passing with OS envar, you can still just `strings /proc/PID/environ` to get the password out

dzou-hpe:
> Does this need a 1.3 PR as well?

plan to create one once this is approved

mharding-hpe (approved):

PR #2233, opened 2022-08-15
CASMINST-5219: Corrections and improvements to install and upgrade documents

mharding-hpe:
# Description

Some of this is minor fixes to issues I found during my recent installs and upgrades of csm-1.2 and csm-1.3.

For csm-1.3, I also updated the CSM health validation document to use the combined Kubernetes/NCN health check script. I removed the needless calls to grep through test totals from there, and from the other places where the changes to the test scripts mean that they are no longer needed.

I also removed some of the Bash scripts that were added to the CSM health validation page, because they duplicated what was already being run in the other tests. They were adding complexity rather than removing it.

# Checklist Before Merging


operations/node_management/Wipe_NCN_Disks.md
```diff
@@ -16,18 +16,20 @@ For wiping Linux on an NCN with a previously installed OS, the [basic wipe](#bas
...
-## Basic Wipe
+## Basic wipe
 
 This wipe erases the magic bits on the disk to prevent them from being recognized and making them ready for deployment, as well as removing the common volume groups.
 
 1. (`ncn#`) List the disks for verification.
 
     ```bash
-    disks_to_wipe="$(lsblk -l -o SIZE,NAME,TYPE,TRAN | grep -E '(raid|'"$metal_transports"')' | sort -u | awk '{print "/dev/"$2}' | tr '\n' ' ' | sed 's/ *$//')"
+    disks_to_wipe=$(lsblk -l -o NAME,TYPE,TRAN | grep -E '[[:space:]].*(sata|nvme|sas|raid)' |
```
rustydb, line 30: Does the grep need the `[[:space:]]`?

Also, I understand that a 1.0 user would use this 1.3 doc for wiping nodes _but_ for 1.3 we don't manually wipe anymore for reinstalls. I think 1.3 and main could still do this:

```bash
. /usr/lib/dracut/modules.d/90metalmdsquash/metal-lib.sh
.. grep -E '(raid|'"$metal_transports"')' ..
```
(where `..` is the rest of the command

rustydb (changes requested):

operations/node_management/Wipe_NCN_Disks.md
mharding-hpe: The `[[:space:]]` part is so that it doesn't match on a device name that happens to have one of those strings in it. I assumed we were only wanting to match strings found in the type or transport fields. Is that not accurate?

For 1.3/main, they may be installing a system that was previously installed with csm-1.0, for example. In that case, that library file will not exist on the system. That's why the wipe page cannot assume it.

rustydb: Yeah I just mean that if a user were to land on this page, it wouldn't be in the context of being on an older image. The upgrade path to 1.3 wouldn't use this, nor would a re-installation or fresh installation. In 1.3 and newer this page is purely operational/trouble-shooting.

For upgrades to anything else (e.g. to 1.2 or 1.0) this page is mandatory and does need to run on an older image.

dzou-hpe (approved):

operations/node_management/Wipe_NCN_Disks.md
mharding-hpe: Sure. I just wasn't sure if we'd rather provide general instructions that will work on a broader range of situations, just in case it's needed, even if that isn't expected. But I'm fine with altering it for 1.3 and main and adding a note about it only working on 1.3+ NCN images, if you prefer. Would we need to update the text that says it will work from the emergency shell in that case? Or would that library be available in that context?

operations/validate_csm_health.md
```diff
@@ -372,7 +348,17 @@ Known issues that may prevent hardware from getting discovered by Hardware State
 
 ## 3 Software Management Services health checks
```
nrockershousen, line 349:
```suggestion
## 3 Software Management Services (SMS) health checks
```

operations/validate_csm_health.md
```diff
@@ -372,7 +348,17 @@ Known issues that may prevent hardware from getting discovered by Hardware State
...
 
-To validate all resources, run [`ncn3.sh`](../scripts/operations/validation/ncn3.sh) script.
+(`ncn-mw#`) To validate all SMS services, run the following:
+
+```bash
+/usr/local/bin/cmsdev test -q all
+```
+
+Successful output end with a line similar to the following:
```
nrockershousen, line 357:
```suggestion
Successful output ends with a line similar to the following:
```

operations/node_management/Wipe_NCN_Disks.md
mharding-hpe: We would definitely need to remove the text that says 
```text
For wiping Linux on an NCN with a previously installed OS, the [basic wipe](#basic-wipe) is sufficient.
```

mharding-hpe: Would also need to change this line
```text
Forcing a node to net-boot that may or may not have a working or desired operating system installed (e.g. a node that might have a non-CSM Linux distro installed)
```

upgrade/Stage_0_Prerequisites.md
```diff
@@ -186,10 +186,11 @@ Stage 0 has several critical procedures which prepare the environment and verify
    If the script does not end with this output, then try rerunning it. If it still fails, see
    [Upgrade Troubleshooting](README.md#relevant-troubleshooting-links-for-upgrade-related-issues).
    If the failure persists, then open a support ticket for guidance before proceeding.
-   
-1. (`ncn-m001#`) Assign a new CFS configuration to the worker node.
 
-The content of the new CFS configuration is described in HPE Cray EX System Software Getting Started Guide S-8000, section "HPE Cray EX Software Upgrade Workflow" subsection "Cray System Management (CSM)". Replace ${NEW_NCN_CONFIGURATION} with the name of the new CFS configuration and ${XNAME} with the component name (xname) of the worker node that was upgraded.
+1. (`ncn-m001#`) Assign a new CFS configuration to the worker nodes.
+
+   The content of the new CFS configuration is described in HPE Cray EX System Software Getting Started Guide S-8000, section
```
nrockershousen, line 192:
```suggestion
   The content of the new CFS configuration is described in *HPE Cray EX System Software Getting Started Guide S-8000*, section
```

bklei (approved):

operations/node_management/Wipe_NCN_Disks.md
rustydb: For the emergency shell they need to source `/lib/metal-lib.sh` instead.

I think we should source the lib for 1.3+ just to remove the duplicated pattern *going forward*. I'll live with the previous releases, and besides we'll never patch one of our previous releases to support more transports.

mharding-hpe: Ok, I'll make the edits. One sec.

nrockershousen (approved):

operations/node_management/Wipe_NCN_Disks.md
mharding-hpe: Done. Let me know what you think.

rustydb: Looks good. nit: it uses `source` for one and `.` for another I'd just make them consistent. Approving this tho.

rustydb (approved):

mharding-hpe:
I had a Slack call with @alexanderkingh to discuss the changes in this PR (specifically removing some of the new CSM health validation scripts), and he approved the changes verbally.

PR #2272, opened 2022-08-18
CASMCMS-8110: Clarify documentation regarding intermediate BOS datastructures

jsl-hpe:
# Description

Addresses customer concern regarding the name of a created configmap for how BOS v1 controller creates intermediate configmaps. This clarification allows customers to perform a more thorough clean-up job for boot sessions launched under BOS v1.

# Checklist Before Merging

### DIFF
--- a/operations/boot_orchestration/Clean_Up_After_a_BOS-BOA_Job_is_Completed_or_Cancelled.md
+++ b/operations/boot_orchestration/Clean_Up_After_a_BOS-BOA_Job_is_Completed_or_Cancelled.md
@@ -2,21 +2,24 @@
 
 When a BOS session is created, there are a number of items created on the system. When a session is cancelled or completed, these items need to be cleaned up to ensure there is not lingering content from the session on the system.
 
-When a session is launched, the items below are created:
+When a session is launched, the following items are created:
 
--   **Boot Orchestration Agent \(BOA\) job:** The Kubernetes job that runs and handles the BOS session.
--   **ConfigMap for BOA:** This ConfigMap contains the configuration information that the BOA job uses. The BOA pod mounts a ConfigMap named `boot-session` at /mnt/boot\_session inside the pod. The name of the configmap has a 1-1 relation to the name of the BOS session created, however, the name of the BOS session can be different from the name of the session template used to create it. For created sessions that do not specify a name, this is most commonly a UUID value.
--   **etcd entries:** BOS makes an entry for the session in its etcd key/value store. If the BOA job has run for long enough, it will also have written a status entry into etcd for this session.
--   **Configuration Framework Service \(CFS\) session:** If configuration is enabled, and the session is doing a boot, reboot, or configure operation, then BOA will have instructed CFS to configure the nodes once they boot. There is not an easy way to link a BOA session to the CFS sessions that are spawned.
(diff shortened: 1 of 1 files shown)


jsl-hpe:
configmap is more appropriate than config map, IMO, but am willing to change it if that makes more sense.

mharding-hpe:
> configmap is more appropriate than config map, IMO, but am willing to change it if that makes more sense.

I think the capitalization is what was throwing it off.

operations/boot_orchestration/Clean_Up_After_a_BOS-BOA_Job_is_Completed_or_Cancelled.md
```diff
@@ -40,140 +43,151 @@ When a session is launched, the items below are created:
...
+                ```
 
-            ```bash
-            cray cfs components update XNAME --desired-state-commit
-            ```
+            1. Prevent the configuration from running.
 
-            This needs to be done for each component. It is enough to prevent configuration from running, and it does not revert to the previous desired state. The previous desired state has already been overwritten at this point, so if the user is trying to completely revert, they will either need to know and apply the previous desired state manually, or return BOS with the previous template using the configure operation \(which may also trigger a configure operation\).
+                ```bash
+                cray cfs components update XNAME --desired-state-commit
```
mharding-hpe, line 119: I don't think this a valid CFS CLI command.

operations/boot_orchestration/Clean_Up_After_a_BOS-BOA_Job_is_Completed_or_Cancelled.md
```diff
@@ -40,140 +43,151 @@ When a session is launched, the items below are created:
...
-            cray cfs components update XNAME --desired-state-commit
-            ```
+                ```bash
+                cray cfs components list
+                ```
+
+            1. Prevent the configuration from running.
+
+                ```bash
+                cray cfs components update XNAME --desired-state-commit
```
mharding-hpe, line 139: Same command here again -- I think this isn't correct. Or anyway, it isn't listed in the CLI help. 

I know this isn't the main purpose of your PR, but we should fix this while we're here.

mharding-hpe (changes requested):
Fix CFS commands.

operations/boot_orchestration/Clean_Up_After_a_BOS-BOA_Job_is_Completed_or_Cancelled.md
rbak-hpe: It's not.  I assume this should be: cray cfs components update XNAME --desired-config ""

mharding-hpe:
```suggestion
                cray cfs components update XNAME --desired-config ""
```

operations/boot_orchestration/Clean_Up_After_a_BOS-BOA_Job_is_Completed_or_Cancelled.md
mharding-hpe:
```suggestion
                cray cfs components update XNAME --desired-config ""
```

mharding-hpe (approved):

PR #2348, opened 2022-08-26
CASMINST-5207: Add typescripts to upgrade procedure & back them up

mharding-hpe:
# Description

This PR adds a step to start a typescript at the beginning of each section of the upgrade procedure, and to stop it at the end of the section. It also makes sure that these are copied over to ncn-m002 when the upgrade moves over to there. A final step is added to the upgrade procedure to back up all of the upgrade log files and typescript files into S3.

I also added a table of contents header to the files which lacked one.

# Checklist Before Merging

### DIFF
--- a/upgrade/README.md
+++ b/upgrade/README.md
@@ -56,7 +56,48 @@ The upgrade is a guided process starting with [Upgrade Management Nodes and CSM 
 - Although it is not recommended, the [Booting CSM `barebones` image](../operations/validate_csm_health.md#5-booting-csm-barebones-image)
   test may be skipped if all compute nodes are active running application workloads.
 
-See [Validate CSM Health](../operations/validate_csm_health.md).
+1. (`ncn-m002#`) If a typescript session is already running in the shell, then first stop it with the `exit` command.
+
+1. (`ncn-m002#`) Start a typescript.
+
+    ```bash
+    script -af /root/csm_upgrade.$(date +%Y%m%d_%H%M%S).post_upgrade_health_validation.txt
+    export PS1='\u@\H \D{%Y-%m-%d} \t \w # '
+    ```
+
+    If additional shells are opened during this procedure, then record those with typescripts as well. When resuming a procedure
+    after a break, always be sure that a typescript is running before proceeding.
+
+1. Validate CSM health.
+
+    See [Validate CSM Health](../operations/validate_csm_health.md).
+
+1. (`ncn-m002#`) Stop typescripts.
+
+    Stop any typescripts that were started during the health validation procedure.
+
+1. (`ncn-m002#`) Backup upgrade logs and typescript files to a safe location.
+
+    1. If any typescript files are on different NCNs, then copy them to `/root` on `ncn-m002`.
+
+    1. Create tar file containing the logs and typescript files.
+
+        > If any typescript file names are not of the form `csm_upgrade.*.txt`, then append their names
+        > to the following `tar` command in order to include them.
+
+        ```bash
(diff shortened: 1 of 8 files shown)


rustydb (approved):

mharding-hpe:
Resolving merge conflicts

upgrade/Stage_2.md
```diff
@@ -17,11 +52,18 @@
 
 1. Repeat the previous step for each other master node **excluding `ncn-m001`**, one at a time.
 
-## Stage 2.2
+## Argo web UI
```
atifsyedali, line 55: Should this just be a bullet point for Stage 2.1 instead of a heading level 2? It kind of breaks the Stage 2.1 --> 2.2 flow.

upgrade/README.md
```diff
@@ -56,7 +56,48 @@ The upgrade is a guided process starting with [Upgrade Management Nodes and CSM
...
+    If additional shells are opened during this procedure, then record those with typescripts as well. When resuming a procedure
+    after a break, always be sure that a typescript is running before proceeding.
+
+1. Validate CSM health.
+
+    See [Validate CSM Health](../operations/validate_csm_health.md).
+
+1. (`ncn-m002#`) Stop typescripts.
+
+    Stop any typescripts that were started during the health validation procedure.
```
atifsyedali, line 77: I am assuming the user already knows how to stop this. Same comment for the rest of the places below where we mention "Stop any typescripts"

upgrade/README.md
```diff
@@ -56,7 +56,48 @@ The upgrade is a guided process starting with [Upgrade Management Nodes and CSM
...
 
-See [Validate CSM Health](../operations/validate_csm_health.md).
+1. (`ncn-m002#`) If a typescript session is already running in the shell, then first stop it with the `exit` command.
+
+1. (`ncn-m002#`) Start a typescript.
+
+    ```bash
+    script -af /root/csm_upgrade.$(date +%Y%m%d_%H%M%S).post_upgrade_health_validation.txt
+    export PS1='\u@\H \D{%Y-%m-%d} \t \w # '
+    ```
```
atifsyedali, line 66: @alexanderkingh Did you want to make this more generic wrt what you were proposing for recording timing information? Perhaps your calls to start / stop a stage can also start / stop typescript? If you are not ready for it yet then nvm. However, I don't think we would want multiple instructions to start / stop logging + metrics.

atifsyedali (approved):
Approving with certain non-blocking comments.

upgrade/README.md
alexanderkingh: Thanks. The name of the typescript at this point is unique enough that we can key off of to record the timing. We'll just match on post_upgrade_health_validation being put into the console. Otherwise this looks fine to me.

upgrade/Stage_2.md
mharding-hpe: Yeah, I agree. The reason I did it this way was to avoid breaking any places where folks may have been referencing stage 2.1, 2.2, etc. But I could have done a docs search for that. Good call out.

upgrade/README.md
mharding-hpe: If we're going to have some kind of "start stage" and "stop stage" thing that we have people run at the beginning/end of stages, then I'd be fine with building in the typescript logic in there. I just wanted to start collecting typescript data for upgrades, since for some reason we never included instructions to do that, even though we always have for installs. It makes debugging issues much easier.

upgrade/README.md
mharding-hpe: Yeah, I went back and forth on that one. I would be fine explicitly including the exit command.

PR #2354, opened 2022-08-29
CASMINST-5241 Added a step to clear any prior enabled CFS sessions from a previous CSM version before creating new CFS sessions for CSM 1.3 content.

alexanderkingh:
# Description

This PR is to address a bug found in [CASMINST-5241](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5241). The bug indicated the issue was CFS pods in a bad state. In reality the bug is more than that. Its possible there are old or bad CFS pods that are continually trying to run that are invalid and are trying to pull from the 1.2 csm-config during the csm 1.3 upgrade. This causes CFS pods to be left in a bad state. This PR attempts to resolve the issue by finding any CFS pods for the management NCNs, and clearing any CFS state until the correct new 1.3 csm-config is uploaded and enabled.

# Checklist Before Merging

### DIFF
--- a/upgrade/Stage_0_Prerequisites.md
+++ b/upgrade/Stage_0_Prerequisites.md
@@ -174,11 +174,25 @@ Stage 0 has several critical procedures which prepare the environment and verify
    [Upgrade Troubleshooting](README.md#relevant-troubleshooting-links-for-upgrade-related-issues).
    If the failure persists, then open a support ticket for guidance before proceeding.
 
-1. (`ncn-m001#`) Before creating a new CFS session for the upgraded content, clear any existing enabled CFS sessions for each management node that might still exist for content from the previous version of CSM.
+1. (`ncn-m001#`) Prevent unexpected CFS sessions from being started on the NCNs during the upgrade.
 
    ```bash
-   for TARGET_XNAME in $(cray hsm state components list --role Management --type Node | grep ^ID  | awk -F\" '{print $2}'); do echo $TARGET_XNAME;curl -s -k -H "Authorization: Bearer ${TOKEN}" -X PATCH "https://api-gw-service-nmn.local/apis/cfs/v2/components/${TARGET_XNAME}" -H 'Content-Type: application/json' -d '{"enabled": true, "state": []}';done
-   ```
+   set -o pipefail
+   if ! XNAMES=$(cray hsm state components list --role Management --type Node --format json |
+                 jq -r '.Components | .[].ID')
+   then
+       echo "ERROR: Unable to generate list of NCN xnames" 1>&2
+   else
+       echo "`echo ${XNAMES} | wc -w` NCN xnames found"
+       ERRORS=0
+       for TARGET_XNAME in ${XNAMES}; do
+           cray cfs components update --state '[]' --enabled true --desired-config '' \
+               "${TARGET_XNAME}" >/dev/null && continue
+           echo "ERROR updating CFS configuration of ${TARGET_XNAME}" 1>&2
+           ERRORS=1
+       done
+       [[ $ERRORS -eq 0 ]] && echo "SUCCESS" || echo "FAILURE" 1>&2
+   fi
 
 1. (`ncn-m001#`) Assign a new CFS configuration to the worker nodes.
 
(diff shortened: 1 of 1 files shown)


upgrade/Stage_0_Prerequisites.md
```diff
@@ -174,6 +174,12 @@ Stage 0 has several critical procedures which prepare the environment and verify
    [Upgrade Troubleshooting](README.md#relevant-troubleshooting-links-for-upgrade-related-issues).
    If the failure persists, then open a support ticket for guidance before proceeding.
 
+1. (`ncn-m001#`) Before creating a new CFS session for the upgraded content, clear any existing enabled CFS sessions for each Management Node that might still exist for content from the previous version of CSM.
```
nrockershousen, line 177:
```suggestion
1. (`ncn-m001#`) Before creating a new CFS session for the upgraded content, clear any existing enabled CFS sessions for each management node that might still exist for content from the previous version of CSM.
```

nrockershousen (approved):

upgrade/Stage_0_Prerequisites.md
```diff
@@ -174,6 +174,12 @@ Stage 0 has several critical procedures which prepare the environment and verify
    [Upgrade Troubleshooting](README.md#relevant-troubleshooting-links-for-upgrade-related-issues).
    If the failure persists, then open a support ticket for guidance before proceeding.
 
+1. (`ncn-m001#`) Before creating a new CFS session for the upgraded content, clear any existing enabled CFS sessions for each management node that might still exist for content from the previous version of CSM.
+
+   ```bash
+   for TARGET_XNAME in $(cray hsm state components list --role Management --type Node | grep ^ID  | awk -F\" '{print $2}'); do echo $TARGET_XNAME;curl -s -k -H "Authorization: Bearer ${TOKEN}" -X PATCH "https://api-gw-service-nmn.local/apis/cfs/v2/components/${TARGET_XNAME}" -H 'Content-Type: application/json' -d '{"enabled": true, "state": []}';done
+   ```
```
mharding-hpe, line 181: This command isn't clearing any CFS sessions -- it's just clearing the CFS state. But that won't prevent CFS from starting sessions. Actually, it will do the opposite -- CFS batcher will immediately start sessions for any NCNs which have a desired configuration set, since their state will no longer match that.

Is the intent here to stop new CFS sessions from being started? If that is the desire, then clearing the desired configuration for the NCNs is what you'd need to do. That would cause CFS batcher to not start any new CFS sessions for the NCNs. I'll make a proposed change here to do that BUT I AM ONLY ASSUMING that this is what you were intending to do. I don't know the full context behind this PR. I just noticed that the language being used did not appear to reflect what was actually being done. In my suggested change I am also leaving what you have (that is, enabling the node and clearing its state). I assume it has been considered that a customer could possibly have one of their NCNs currently disabled in CFS, and this is deliberately enabling all of them regardless of that. If that has not been considered, then it should be. 

It also looks kind of funny having the mismatch of using the CLI for HSM but the API for CFS, IMO.

Finally, the CLI command appears to be assuming that the output format will be TOML, but it does not specify that. We should **always** explicitly specify output format when using the CLI. In this case, I'm suggesting a change to use JSON and jq, so it can be more precise about getting the desired information from the output.

````suggestion
1. (`ncn-m001#`) Prevent unexpected CFS sessions from being started on the NCNs during the upgrade.

   ```bash
   set -o pipefail
   if ! XNAMES=$(cray hsm state components list --role Management --type Node --format json |
                 jq -r '.Components | .[].ID')
   then
       echo "ERROR: Unable to generate list of NCN xnames" 1>&2
   else
       echo "$(echo ${XNAMES} | wc -w) NCN xnames found"
       ERRORS=0
       for TARGET_XNAME in ${XNAMES}; do
           cray cfs components update --state '[]' --enabled true --desired-config '' \
               "${TARGET_XNAME}" >/dev/null && continue
           echo "ERROR updating CFS configuration of ${TARGET_XNAME}" 1>&2
           ERRORS=1
       done
       [[ $ERRORS -eq 0 ]] && echo "SUCCESS" || echo "FAILURE" 1>&2
   fi
   ```
````

mharding-hpe (changes requested):

rustydb (approved):

mharding-hpe:
I will be creating a new PR for this.

PR #2381, opened 2022-08-31
CASMPET-5846

mitchty:
# Description

Add initial docs for kubernetes encryption setup and usage. As well as add the script used to setup the encryption after install or upgrade.

As this script was largely unit tested for another repo bring in similar infrastructure for shellspec to keep the unit test framework in place for future updates or bugfixes.

This was tested largely on redbull where these examples come from, though I did scrub dates in example output.  I have validated on other systems before rebuild but there isn't much difference.

# Checklist Before Merging

- [x ] If I added any command snippets, the steps they belong to follow the prompt conventions (see [example][1]).
- [x ] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].
- [x ] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -43,6 +43,8 @@ shown here with numbered topics.
     1. [Prepare compute nodes](#9-prepare-compute-nodes)
     1. [Next topic](#10-next-topic)
     - [Troubleshooting installation problems](#troubleshooting-installation-problems)
+1. [Post-installation](#post-installation)
+    1. [Kubernetes encryption](#11-encryption)
 
 > **`NOTE`** If problems are encountered during the installation,
 > [Troubleshooting installation problems](#troubleshooting-installation-problems) and
@@ -200,3 +202,11 @@ switches for the HPE Cray EX system. The procedures in this section should be re
 for additional information on system hardware, troubleshooting, and administrative tasks related to CSM.
 
 See [Troubleshooting Installation Problems](troubleshooting_installation.md).
+
+### 11. Kubernetes encryption
+
+As an optional post installation task you may enable encryption of `kubernetes`
+secrets. This enables at rest encryption of data in the `etcd` database used by
+kubernetes.
+
+See [Kubernetes Encryption](../operations/encryption/Overview.md)
--- a/install/README.md
+++ b/install/README.md
@@ -205,8 +205,7 @@ See [Troubleshooting Installation Problems](troubleshooting_installation.md).
 
 ### 11. Kubernetes encryption
 
-As an optional post installation task you may enable encryption of Kubernetes
-secrets. This enables at rest encryption of data in the `etcd` database used by
-Kubernetes.
+As an optional post installation task, encryption of Kubernetes secrets may be enabled. This enables
+at rest encryption of data in the `etcd` database used by Kubernetes.
 
 See [Kubernetes Encryption](../operations/encryption/Overview.md)
--- a/install/README.md
+++ b/install/README.md
@@ -208,4 +208,4 @@ See [Troubleshooting Installation Problems](troubleshooting_installation.md).
 As an optional post installation task, encryption of Kubernetes secrets may be enabled. This enables
 at rest encryption of data in the `etcd` database used by Kubernetes.
 
-See [Kubernetes Encryption](../operations/encryption/Overview.md).
+See [Kubernetes Encryption](../operations/kubernetes/encryption/README.md).
--- a/install/README.md
+++ b/install/README.md
@@ -205,8 +205,8 @@ See [Troubleshooting Installation Problems](troubleshooting_installation.md).
 
 ### 11. Kubernetes encryption
 
-As an optional post installation task you may enable encryption of `kubernetes`
+As an optional post installation task you may enable encryption of Kubernetes
 secrets. This enables at rest encryption of data in the `etcd` database used by
-kubernetes.
+Kubernetes.
 
 See [Kubernetes Encryption](../operations/encryption/Overview.md)
--- a/install/README.md
+++ b/install/README.md
@@ -208,4 +208,4 @@ See [Troubleshooting Installation Problems](troubleshooting_installation.md).
 As an optional post installation task, encryption of Kubernetes secrets may be enabled. This enables
 at rest encryption of data in the `etcd` database used by Kubernetes.
 
-See [Kubernetes Encryption](../operations/encryption/Overview.md)
+See [Kubernetes Encryption](../operations/encryption/Overview.md).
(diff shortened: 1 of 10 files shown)


operations/encryption/Overview.md
```diff
@@ -0,0 +1,160 @@
...
+
+If for any reason you wish to force a rewrite of secret data, you may do so by overwriting the annotation used by `cray-k8s-encryption`
+
+* (`ncn-mw#`) Force a rewrite of existing data:
+
+    ```bash
+    kubectl annotate secret --namespace kube-system cray-k8s-encryption current=rewrite --overwrite
+    ```
+
+    This command gives no output on succes.
```
mitchty, line 160: Gah how did spellcheck not find this.

kimjensen-hpe (approved):

install/README.md
```diff
@@ -200,3 +202,11 @@ switches for the HPE Cray EX system. The procedures in this section should be re
 for additional information on system hardware, troubleshooting, and administrative tasks related to CSM.
 
 See [Troubleshooting Installation Problems](troubleshooting_installation.md).
+
+### 11. Kubernetes encryption
+
+As an optional post installation task you may enable encryption of Kubernetes
+secrets. This enables at rest encryption of data in the `etcd` database used by
+Kubernetes.
```
mharding-hpe, line 210:
```suggestion
As an optional post installation task, encryption of Kubernetes secrets may be enabled. This enables
at rest encryption of data in the `etcd` database used by Kubernetes.
```

install/README.md
```diff
@@ -200,3 +202,11 @@ switches for the HPE Cray EX system. The procedures in this section should be re
...
+
+See [Kubernetes Encryption](../operations/encryption/Overview.md)
```
mharding-hpe, line 212:
```suggestion
See [Kubernetes Encryption](../operations/encryption/Overview.md).
```

operations/encryption/Overview.md
```diff
@@ -0,0 +1,160 @@
+# Kubernetes Encryption
+
+Beginning in CSM 1.3, support is enabled for data encryption in `etcd` for Kubernetes secrets at rest.
+
+This controller is deployed by default in CSM with the `cray-kubernetes-encryption` Helm chart.
+
+By default encryption is not enabled and must be enabled after install.
```
mharding-hpe, line 7:
```suggestion
By default, encryption is not enabled and must be enabled after install.
```

mharding-hpe:
Adding "do not merge" label because I want to review this before we merge it (and have had bad luck with things being merged when I have only commented on this in my feedback but not added the label). This does not imply anything about my feelings as to the quality of the content here -- it just means that I don't know if I will have time to review this before the first time it will be considered for merging.

I will remove the label once I've finished my reviews and issues identified (if any) are addressed.

mitchty:
Ok thanks for the quick review @mharding-hpe the biggest questionable bit here might be the shellspec unit test library but it is a separate commit so I can easily remove it. But I really want to have more unit tests of our shell in general and wanted to lead with examples. It was a part of a commit I didn't get in in time for the node images but I think in general our shell could use a few more unit tests.

scripts/operations/node_management/encryption.sh
```diff
@@ -0,0 +1,405 @@
...
+validprovider() {
+  provider=${1:-invalid}
+  if [ "aesgcm" = "${provider}" ] || [ "aescbc" = "${provider}" ]; then
+    return 1
+  fi
+  return 0
+}
+
+# Validate a secret type, identity is a special case
+validate() {
```
mharding-hpe, line 61: I think I must be missing something here. I don't see this function ever being called in this script, and it seems odd to make a function which just returns 1. It feels like either:
1. This script sources some other shell file, and that file calls this function
2. Some other script sources THIS script, and that file calls this function.

But I don't see this script sourcing anything else. I didn't check to see if the other scripts in this PR sourced this file, but I do see that they never call the validate function that I could see. 

What am I missing?

operations/encryption/Overview.md
```diff
@@ -0,0 +1,160 @@
+# Kubernetes Encryption
```
mharding-hpe, line 1: If this is specifically for Kubernetes encryption, would it make more sense for this to be located in the `operations/kubernetes/encryption/` folder? And usually for the main file on a topic, we would name it README.md (and create a symlink for index.md pointing to it).

Alternatively, if we want this in the general encryption folder because we anticipate grouping it with other, non-k8s encryption topics, then our convention is to have the file name match its main header. So in this case, it would be Kubernetes_Encryption.md.

mharding-hpe (changes requested):

mharding-hpe:
> Ok thanks for the quick review @mharding-hpe the biggest questionable bit here might be the shellspec unit test library but it is a separate commit so I can easily remove it. But I really want to have more unit tests of our shell in general and wanted to lead with examples. It was a part of a commit I didn't get in in time for the node images but I think in general our shell could use a few more unit tests.

You will never hear me argue against more tests, believe me. :)

operations/encryption/Overview.md
mitchty: I can change it to whatever makes sense. I'll swap it out for the latter. I was mostly following some other conventions I had seen.

scripts/operations/node_management/encryption.sh
mitchty: Probably leftovers from a red/green refactor, I'll nuke it.

mitchty:
Ok I swapped the files into operations/kubernetes/encryption and scripts/operations/kubernetes respectively. Sat on it overnight but this seems a better layout with your suggestions. I don't think this will ever expand out beyond just kubernetes encryption in etcd really.

operations/kubernetes/encryption/README.md
```diff
@@ -0,0 +1,157 @@
...
+Beginning in CSM 1.3, support is enabled for data encryption in `etcd` for Kubernetes secrets at rest.
+
+This controller is deployed by default in CSM with the `cray-kubernetes-encryption` Helm chart.
+
+By default, encryption is not enabled and must be enabled after install.
+
+## Table of contents
+
+* [Implementation details](#implementation-details)
+* [Enabling encryption](#enabling-encrypion)
```
rgross727, line 12: [Enabling encryption](#enabling-encrypion)
spelling on encryption

rgross727: Fixed now.

mharding-hpe (commented):
My previous concerns appear to be resolved

mharding-hpe (approved):
I haven't had a chance to fully re-review all of the changes since the last time I reviewed it, but I did confirm that my original concerns appear to be addressed.

rustydb (approved):

PR #2398, opened 2022-09-01
CASMINST-5298: Deploy Helm charts properly in prerequisites script

mharding-hpe:
# Description

During the CSM upgrade, in stage 0, the prerequisites.sh script deploys three helm charts ahead of when it normally would do so. However, the code to do this currently does not base its charts on the overall CSM manifest files. At least in the case of csm-config, this makes a difference. It also does not customize any of the charts, although it's less clear if that is a big problem. In any case, this PR modifies the prerequisites script so that these 3 charts (csm-config, cray-kyverno, and kyverno-policy) are deployed based on their CSM manifest entries, after being customized.

I tested this on mug and it works, but I had to jump through some hoops because of the bad state that mug is in. I'm going to open this for review, but I'd really like to be able to test this on an actual 1.2 system. I feel confident in the change, but would feel more confident after proper testing.

On mug, here is the customized csm-config manifest file that ended up getting deployed:
```yaml
apiVersion: manifests/v1beta1
metadata:
  name: csm-config
spec:
  charts:
  - name: csm-config
    namespace: services
    source: csm
    values:
      cray-import-config:
        catalog:
          image:
            tag: 1.3.1
        import_job:
          CF_IMPORT_GITEA_REPO: csm-config-management
          CF_IMPORT_PRODUCT_NAME: csm
          CF_IMPORT_PRODUCT_VERSION: 1.3.0-beta.51
    version: 1.12.0
  sources:
    charts:
    - location: /etc/cray/upgrade/csm/csm-1.3.0-beta.51/tarball/csm-1.3.0-beta.51/helm
      name: csm
      type: directory
```

As you can see, it contains the stanzas that had previously been missing. And I verified that it correctly launched the csm-config-import job. That job completed in under 2 minutes, and the prerequisites script will take longer than that to complete, so I don't think we need to explicitly wait for it to complete in the script. But that's an option we can add, if desired.

# Checklist Before Merging

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -30,7 +30,7 @@ locOfScript=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 . ${locOfScript}/../common/ncn-common.sh $(hostname)
 trap 'err_report' ERR INT TERM HUP EXIT
 # array for paths to unmount after chrooting images
-#shellcheck disable=SC2034
+# shellcheck disable=SC2034
 declare -a UNMOUNTS=()
 
 while [[ $# -gt 0 ]]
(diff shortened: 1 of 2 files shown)


rustydb (approved):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -194,9 +230,9 @@ if [[ $state_recorded == "0" ]]; then
         # copy the NTP script and template to the target ncn
         rsync -aq "${CSM_ARTI_DIR}"/chrony "$target_ncn":/srv/cray/scripts/common/
 
-        # shellcheck disable=SC2029 # it's ok that $TOKEN expands on the client side
+        # shellcheck disable=SC2029 # it is intentional that ${TOKEN} expands on the client side
         # run the script
-        if ! ssh "$target_ncn" "TOKEN=$TOKEN /srv/cray/scripts/common/chrony/csm_ntp.py"; then
+        if ! ssh "$target_ncn" "TOKEN=${TOKEN} /srv/cray/scripts/common/chrony/csm_ntp.py"; then
```
mharding-hpe, line 235:
```suggestion
        if ! ssh "${target_ncn}" "TOKEN=${TOKEN} /srv/cray/scripts/common/chrony/csm_ntp.py"; then
```

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -391,28 +427,28 @@ state_recorded=$(is_state_recorded "${state_name}" "$(hostname)")
...
+    "${EXTRACT_CHART_MANIFEST}" csm-config "${CSM_MANIFESTS_DIR}/sysmgmt.yaml" > "${TMP_MANIFEST}"
+    cat "${TMP_MANIFEST}"
+
+    # Customize it
+    TMP_MANIFEST_CUSTOMIZED=$(mktemp --tmpdir="${manifest_folder}" csm-config.customized.XXXXXX.yaml)
+    echo "${TMP_MANIFEST_CUSTOMIZED}"
+    manifestgen -i "${TMP_MANIFEST}" -c "${TMP_CUST_YAML}" -o "${TMP_MANIFEST_CUSTOMIZED}"
+    cat "${TMP_MANIFEST_CUSTOMIZED}"
+
+    loftsman ship --manifest-path "${TMP_MANIFEST_CUSTOMIZED}"
```
mharding-hpe, line 451:
```suggestion
    loftsman ship --manifest-path "${TMP_MANIFEST_CUSTOMIZED}"

```

PR #2416, opened 2022-09-06
CASMINST-5088 : Documentation for troubleshooting kyverno

pradeepkumargl:
# Description

Kyverno is a new introduction for this release. Incase customer encounters any issue the steps mentioned here helps for debugging manually.

### DIFF
--- a/operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
+++ b/operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
@@ -22,4 +22,4 @@ This is used to check the Kyverno policy report for any Failures, Warnings, Erro
 
 ### For more information check:
 
-* [Kyverno](../operations/kubernetes/Kyverno.md)
+* [Kyverno](../kubernetes/Kyverno.md)
--- a/operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
+++ b/operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
@@ -1,6 +1,6 @@
-# How to troubleshoot Kyverno configuration manually:
+# How to troubleshoot Kyverno configuration manually
 
-## Check if Kyverno pods are up and running.
+## Check if Kyverno pods are up and running
 
 Run the script with -p option.
 
@@ -10,7 +10,7 @@ Run the script with -p option.
 
 This will list the expected Kyverno pods in running state.
 
-## Check if Kyverno policy report doesn’t have Failures, Warnings, Errors and Skipped policies count.
+## Check if Kyverno policy report doesn’t have Failures, Warnings, Errors and Skipped policies count
 
 Run the script with -p option.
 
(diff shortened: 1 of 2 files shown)


rustydb (approved):

operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
```diff
@@ -0,0 +1,25 @@
+# How to troubleshoot Kyverno configuration manually
+
+## Check if Kyverno pods are up and running
+
+Run the script with -p option.
+
+```bash
+./k8s_kyverno_pods_running.sh -p
```
djlapenta, line 8: Hi Pradeep - where would the user get the script if they wanted to use it for troubleshooting - where does it get installed on the system?   We should show the whole pathway.

djlapenta (changes requested):

operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Check if Kyverno policy report doesn’t have Failures, Warnings, Errors and Skipped policies count
+
+Run the script with -p option.
+
+```bash
+./k8s_kyverno_polr_list.sh -p
+```
+
+This is used to check the Kyverno policy report for any Failures, Warnings, Errors and Skipped policies count.
```
djlapenta, line 21: Is there any guidance to give the user in the event that they do have a "failure" case on the test?  When we move it to a goss test, we will need to include something in the error message to tell the user where to get more info so that their issue can be resolved and/or how to further troubleshoot it.  Is the kyverno link below all that is needed to help with troubleshooting a non-passing policy report?

operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
djlapenta: I think we want to give them the pathway to find these scripts so that if they need to run them, they are not searching for them on the system..  I dropped the path in the slack channel.

djlapenta (approved):

operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Check if Kyverno pods are up and running
+
+Run the script with -p option.
+
+```bash
+/opt/cray/tests/install/livecd/scripts/k8s_kyverno_pods_running.sh -p
+```
+
+This will list the expected Kyverno pods in running state.
```
jeremy-duckworth, line 11: Running this command on a system with 1.3.0-beta.74 does not show Kyverno pods in a running state, it only prints `PASS`. This irrespective of whether or not the `-p` is provided. The documentation should be clear on expected output (e.g., the only time `PASS` isn't printed is when an error exists).

jeremy-duckworth (changes requested):

operations/security_and_authentication/Troubleshoot_Kyverno_Configuration_manually.md
rustydb: Stale comment, marking as closed.

PR #2426, opened 2022-09-06
Shell shenanigans

mitchty:
# Description

This is a POC branch to add shfmt github actions on merge and an editorconfig. As I couldn't find "standards" I'm just running with the defaults that the editorconfig community uses and it appears based on the shell that already is present is using. Which means 2 spaces no tabs. If we want to change that we can.

If you want to run this locally you may by using:
https://github.com/nektos/act

And running *act -j sh-checker*

Note I am using the default editorconfig from here:
https://github.com/editorconfig/editorconfig-defaults/blob/master/.editorconfig

It is setup so we don't need to muck with yaml files.

shfmt is this tool: https://github.com/mvdan/sh

As well as I am using the following for shell format as that changes the least amount of shell in this repo and makes the most sense in my opinion:

shfmt -ci -p -i 2 -bn -sr

Each switch in turn is:
- -ci, --case-indent Switch cases will be indented.
- -p, --posix Shorthand for -ln=posix. (in spite of all our bash this helps change as little as possible, I'd prefer .sh be posix and .bash be actual bash scripts so we can be a bit better about this but its not critical)
- -i, --indent Indent: 0 for tabs (default), >0 for number of spaces. (the vast majority of our scripts are already this, most of our changes are invalid indentation that doesn't follow this so just running with it)
- -bn, --binary-next-line Binary ops like && and | may start a line. (only a few scripts seem to violate this one, so seems like this is already our standard)
- -sr, --space-redirects Redirect operators will be followed by a space. (seems the right thing to be honest)

I also setup the action for now to comment on the changes in the pr. We can turn this off when we want to merge.

The blocker I can see for now to not use this is that I will likely need to add support to the action to only check changed files.

What I want to do is submit a PR for this line https://github.com/luizm/action-sh-checker/blob/master/entrypoint.sh#L39 to allow for an env var to pass in the files to be checked. For now this checks everything but it runs a lot more than just shellcheck.

We could fork this action in the interim to accomplish that. There is an upstream issue for that specific use case but I think letting the action specify what it wants checked via an env var is probably better as it allows for more control than just updated xyz. 

Upstream pr https://github.com/luizm/action-sh-checker/issues/27

# Checklist Before Merging

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,10 +1,13 @@
 root = true
 
 [*]
-end_of_line = lf
-charset = utf-8
-
-[*.sh]
-charset = utf-8
 indent_style = space
 indent_size = 2
+end_of_line = lf
+trim_trailing_spaces = true
+insert_final_newline = true
+
+# trailing spaces in markdown indicate word wrap
+[*.md]
+trim_trailing_spaces = false
+max_line_length = 80
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+
+[*.sh]
+charset = utf-8
+indent_style = space
+indent_size = 2
--- a/.github/workflows/sh-checker.yaml
+++ b/.github/workflows/sh-checker.yaml
@@ -26,21 +26,23 @@ on:
   pull_request:
 
 jobs:
-  shell-check:
+  sh-checker:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v18.7
+        with:
+          files: "**/*.sh"
+          files_ignore: ".github/**/*"
 
-    - name: Get changed files
-      id: changed-files
-      uses: tj-actions/changed-files@v18.7
-      with:
-        files: "**/*.sh"
-        files_ignore: ".github/**/*"
-
-    - uses: docker://koalaman/shellcheck:stable
-      if: ${{ steps.changed-files.outputs.all_changed_files }}
-      id: shell-check
-      with:
-        args: --severity=warning --color=always ${{ steps.changed-files.outputs.all_changed_files }}
-
+      - name: Run the sh-checker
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
```
rustydb, line 5: I think we want `4`.
Only in the case of unordered/ordered lists do we want `2`..

mitchty: If we do that we'll need to add a block for yaml/yml files as the indentation requirement there is 2 spaces no tabs:
https://docs.saltproject.io/en/latest/topics/troubleshooting/yaml_idiosyncrasies.html

As well as consider if shell needs to have a different indent style/setting as well. Part of why i just ran with these defaults is it covers a lot of this "for free". Let me see what the shell looks like change wise with 4 spaces. I think it'll change a lot more loc.

mitchty: A quick and very rough pass on 4 spaces for sh vs 2 (just looking for hunk +/1 lines it will catch a bit more than that though so take with salt):

4 spaces:
```
$ act -j sh-checker stdboth | grep -Ec '\|\ [-+]'
949
zsh: exit 1     act -j sh-checker 2>&1 | 
zsh: done       grep -Ec '\|\ [-+]'
```

2 spaces
```
$ act -j sh-checker stdboth | grep -Ec '\|\ [-+]' 
461
zsh: exit 1     act -j sh-checker 2>&1 | 
zsh: done       grep -Ec '\|\ [-+]'
```

I didn't do a comprehensive check but tabs is even more changes as it touches every indentation line.

So if we define this or want to define it we need to define:

Default for everything (my thought here is this should be the thing that most stuff needs)
Default for specific file types if it doesn't fit the above. Aka for adoc maybe we want tabs and 8 spaces, we would define a block for [*.adoc]

I didn't look for Makefiles but we may want a block for those too to specifiy tab and whatever tab stop we want for them

rustydb: Sorry for the delay.

I think we can ignore adoc for now.

Makefiles could use tabs.

Am I right in understanding that we're okay with using 4 spaces as a default?

nrockershousen (approved):

PR #2459, opened 2022-09-14
CASMINST-5327 war var that is not set yet

dzou-hpe:
# Description

# Checklist Before Merging

### DIFF
--- a/upgrade/scripts/upgrade/prepare-assets.sh
+++ b/upgrade/scripts/upgrade/prepare-assets.sh
@@ -26,6 +26,7 @@
 set -e
 locOfScript=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 . ${locOfScript}/../common/upgrade-state.sh
+CSM_ARTI_DIR="not_set"
 #shellcheck disable=SC2046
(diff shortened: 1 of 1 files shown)


bklei (approved):

upgrade/scripts/upgrade/prepare-assets.sh
```diff
@@ -26,6 +26,7 @@
 set -e
 locOfScript=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 . ${locOfScript}/../common/upgrade-state.sh
+CSM_ARTI_DIR="not_set"
```
mharding-hpe, line 29: Why not instead set the variable correctly and then move the call to include ncn-common.sh until after this variable is set? I'll make a PR to your PR to show you what I mean.

mharding-hpe (changes requested):

upgrade/scripts/upgrade/prepare-assets.sh
mharding-hpe: See what you think of this: https://github.com/Cray-HPE/docs-csm/pull/2460

I think the dummy value would work, but it seems cleaner to me to just set the correct value

mharding-hpe (approved):

PR #2495, opened 2022-09-16
CASMINST-5347: Add retries and additional output to some upgrade workflows; Remove unused script

mharding-hpe:
# Description

While trying to investigate [CASMINST-5330](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5330), I noticed that some parts of the upgrade workflows would be improved by adding additional output and also retries on things which may fail due to transient issues. This PR adds those, and also removes a script which is no longer used in 1.3, but remained in the repo.

I did some testing of the updates on surtur.

main backport: https://github.com/Cray-HPE/docs-csm/pull/2503

# Checklist Before Merging


upgrade/scripts/upgrade/prepare-assets.sh
```diff
@@ -81,6 +79,9 @@ fi
 
 CSM_REL_NAME="csm-${CSM_RELEASE}"
 CSM_ARTI_DIR="/etc/cray/upgrade/csm/${CSM_REL_NAME}/tarball/${CSM_REL_NAME}"
+
+locOfScript=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+. "${locOfScript}/../common/upgrade-state.sh"
```
leliasen-hpe, line 84: I don't think this line is necessary

mharding-hpe: I just moved it down from earlier in the file. I assumed it was there because it was needed, but if it isn't, I can remove it.

leliasen-hpe: You are right that it is needed. But line 85 does the same think I think and provides a specific node. This is line 85 `. "${locOfScript}/../common/ncn-common.sh" "$(hostname)"`. I don't think you would need both of 84 and 85, correct?

mharding-hpe: Oh, I didn't realize that ncn-common.sh also included upgrade-state. I see that it does, though. Ok, I can remove that line. Happy to eliminate redundancy.

leliasen-hpe: Wait, I was very wrong, sorry. You were right initially. I that they were the same file for some reason

leliasen-hpe (approved):

rustydb (approved):

PR #2496, opened 2022-09-20
Apply consistency to triple-backtick markup

pawsey-kbuckley:
# Description

This file
   - `operations/argo/Using_Argo_Workflows.md`

uses triple backticks in the running text to render command flag
arguments into teletype text.

Everywhere else (yes, I grep-ed it) within the documentation source,
just using a single backtick achieves the same thing, for example, in
   - `operations/configure_cray_cli.md`

we read

```
At this point, the script may be re-run with the `--debug` flag added,
```

which renders as

> At this point, the script may be re-run with the `--debug` flag added,

The triple ticks, whilst not being incorrect, hinder attempts to
automatically handle triple-back-ticked code blocks that contain
characters that the Jekyll renderer considers special.

Any chance of some consistency in the running text sections?

This PR, should you choose to acceot it, applies some consistency to the file in question.

### DIFF
--- a/operations/argo/Using_Argo_Workflows.md
+++ b/operations/argo/Using_Argo_Workflows.md
@@ -11,28 +11,28 @@ This page provides information on operating Argo workflows in CSM and describes 
 * If any step fails, by default, that step will be continuously retried until it succeeds.
 There are two ways to make Argo not continuously retry a failed step.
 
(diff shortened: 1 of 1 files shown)


rustydb (approved):

rustydb:
Thank you.

nrockershousen (approved):

PR #2504, opened 2022-09-20
CASMCMS-8236: Update documentation for image customization, including examples

rbak-hpe:
# Description

This updates the recommendations for how to write Ansible playbooks for image customization, including adding sample templates that can be used as a starting point.

# Checklist Before Merging

### DIFF
--- a/operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
+++ b/operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
@@ -3,24 +3,59 @@
 The Configuration Framework Service \(CFS\) enables Ansible playbooks to run against both running nodes \(node personalization\) and images prior to boot\(image customization\).
 See [Configuration Management Use Cases](Configuration_Management.md#use-cases) for more information about image customization and when it should be used.
 
-Ideally image customization playbooks should be separate from node personalization playbooks. This reduces the likelihood that Ansible content will run in both modes, and reduces the need for conditional checks in the playbooks.
+## Using the `cfs_image` host group
 
-For situations when both image customization and node personalization need to be part of the same playbook, CFS provides the `cray_cfs_image` variable to distinguish between
-node personalization and image customization. When this variable is set to `true`, it indicates that the CFS session is an image customization and the playbook is targeting an image.
+During image customization, CFS will automatically add all image customization hosts into a special `cfs_image` host group in Ansible inventory.
+Plays intended for image customization can then target this group in addition to any other provided host groups.
 
-## Using `cray_cfs_image`
+To target only image customization, plays should use the following syntax.  In this example the play is targeting only _images_ for `Compute` nodes. `&` takes the intersection of the `Compute` and `cfs_image` groups.
 
-`cray_cfs_image` can be set with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
+```yaml
+hosts: Compute:&cfs_image
+```
+
+To target only node personalization, plays should use the following syntax.  In this example the play is targeting only _running_ `Compute` nodes. `!` negates the `cfs_image` group, so that only Compute nodes that are not an image are targeted.
+
+```yaml
+hosts: Compute:!cfs_image
+```
+
+For more information on complex host targets, see the [Ansible Hosts Documentation](https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html).
+
+### Example: Using the `cfs_image` host group for image customization
+
+```yaml
+- name: Image customization play
+  hosts: Management_Worker:&cfs_image
+  tasks:
+    - include_role:
+        role: cos-services-install
+
+- name: Node personalization play  
+  hosts: Management_Worker:!cfs_image
+  tasks:
+    - include_role: 
+        role: cos-services-restart
+```
+
+## Using the `cray_cfs_image` variable
+
+`** NOTE **` This option is no longer recommended and should only be used in small playbooks and one-off cases
+ as it is more efficient for Ansible to determine this at the host level rather than checking the `cray_cfs_image` variable for multiple tasks.
+ The preferred method is to use the aforementioned `cfs_image` host group.
+
+CFS also provides the `cray_cfs_image` variable to distinguish between node personalization and image customization.
+`cray_cfs_image` can be used with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
 to selectively run individual tasks with `when: cray_cfs_image`, or to ignore individual tasks with `when: not cray_cfs_image`.
 
-Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#applying-when-to-roles-imports-and-includes)\).
+Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#conditionals-with-includes).
 In instances where the same playbook may be run in both modes, it is best practice to include a conditional on all parts of the playbook. This is best done by placing the conditional on an `include_*` statement.
 See [Write Ansible Code for CFS: Reduce wasted time](Write_Ansible_Code_for_CFS.md#reduce-wasted-time) for more information on optimizing conditionals.
 
-It is also best practice to include a default in Ansible roles for playbook and role portability because CFS injects this variable at runtime. This can be done in the defaults section of the role, or where the variable is called. For example:
+It is best practice to include a default in Ansible roles for playbook and role portability because CFS injects this variable at runtime. This can be done in the defaults section of the role, or where the variable is called. For example:
 
-```yaml
-when: "{{ cray_cfs_image | default(false) }}"
+```text
+{{ cray_cfs_image | default(false) }}
 ```
 
 If a default is not provided, any playbooks or roles will not be runnable outside of the CFS Ansible Execution Environment \(AEE\) without the user specifying `cray_cfs_image` in the `vars` files or with the Ansible `extra-vars` options.
--- a/operations/configuration_management/Write_Ansible_Code_for_CFS.md
+++ b/operations/configuration_management/Write_Ansible_Code_for_CFS.md
@@ -12,7 +12,7 @@ Help with Ansible can be found in the external Ansible documentation:
 
 ## Ansible Code Structure
 
-The Version Control Service \(VCS\) is setup during the Cray System Management \(CSM\) product installation, which is the appropriate place to store configuration content.
+The Version Control Service \(VCS\), which is setup during the Cray System Management \(CSM\) product installation, is the appropriate place to store configuration content.
 Individual product installations include the Ansible code to properly configure each product.
 
 The structure of the individual repository directories matches the [recommended directory layout](https://docs.ansible.com/ansible/2.9/user_guide/playbooks_best_practices.html#content-organization) from the external Ansible documentation.
@@ -27,7 +27,7 @@ For example, a playbook may contain a list of roles that are run on only the `Co
 For example, `hosts: Compute` would be used to target the compute nodes. Users can create additional sections that target other node types, or adjust the hosts that the included roles will run against as necessary.
 It is also possible to target multiple groups within a section of a playbook, or to specify complex targets, such as nodes that are in one group and not in another group.
 The syntax for this is available in the external [Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html#common-patterns).
-Hosts can also be in more than one group at a time. In this case, Ansible will run all sections that match the node type against the node.
+Hosts can also be in more than one group at a time. In this case, Ansible will run all sections where the node is in the targeted host group.
 
 See the [Ansible Inventory](Ansible_Inventory.md) section for more information about groups that are made automatically available through CFS dynamic inventory.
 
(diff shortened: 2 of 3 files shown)


operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
-`cray_cfs_image` can be set with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
-to selectively run individual tasks with `when: cray_cfs_image`, or to ignore individual tasks with `when: not cray_cfs_image`.
+```text
+{{ cray_cfs_image | default(false) }}
+```
 
-Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#applying-when-to-roles-imports-and-includes)\).
-In instances where the same playbook may be run in both modes, it is best practice to include a conditional on all parts of the playbook. This is best done by placing the conditional on an `include_*` statement.
-See [Write Ansible Code for CFS: Reduce wasted time](Write_Ansible_Code_for_CFS.md#reduce-wasted-time) for more information on optimizing conditionals.
+If a default is not provided, any playbooks or roles will not be runnable outside of the CFS Ansible Execution Environment \(AEE\) without the user specifying `cray_cfs_image` in the `vars` files or with the Ansible `extra-vars` options.
```
jsollom-hpe, line 17: This is an excellent call out.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+
+CFS automatically sets this variable in the `hosts/01-cfs-generated.yaml` file for all sessions. When the session target is image customization, it sets `cray_cfs_image` to `true`; otherwise, it is `false`.
+
+### Using `cray_cfs_image` with `group_by`
+
+`cray_cfs_image` can be used at the top of the playbook with Ansible's `group_by` module.
+This turns the variable into a host group, and allows Ansible to more efficiently exclude tasks from being run, rather than checking a conditional for every task.
```
jsollom-hpe, line 24: Is it worth pointing out that the syntax all:&image_true is the intersection of all hosts and those hosts in the image_true group? This may be a common Ansible host pattern, but I was not familiar with it and had to look it up.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+- import_playbook: playbook2.yaml
 ```
 
-If a default is not provided, any playbooks or roles will not be runnable outside of the CFS Ansible Execution Environment \(AEE\) without the user specifying `cray_cfs_image` in the `vars` files or with the Ansible `extra-vars` options.
+See Ansible's [playbooks best practices](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#handling-os-and-distro-differences) guide for more information on `group_by`
 
-CFS automatically sets this variable in the `hosts/01-cfs-generated.yaml` file for all sessions. When the session target is image customization, it sets `cray_cfs_image` to `true`; otherwise, it is `false`.
+### Using `cray_cfs_image` with conditionals
+
+When `group_by` cannot be used, `cray_cfs_image` can be set with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
```
haasken-hpe, line 69: minor suggestion:

```suggestion
When `group_by` cannot be used, `cray_cfs_image` can be used with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
```

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+# Node personalization
+- name: Node personalization
+  hosts: all:&image_false
+  gather_facts: false
+  tasks:
+    ...
+```
 
-It is also best practice to include a default in Ansible roles for playbook and role portability because CFS injects this variable at runtime. This can be done in the defaults section of the role, or where the variable is called. For example:
+When multiple playbooks are imported from a top level playbook, the `group_by` task can also be moved to the top  level playbook so that it only needs to run once
```
jsollom-hpe, line 50: Moving the group_by task to the top level is more efficient, but it renders the imported playbooks without a definition of those image_{true,false} groups if they are run outside of the top level playbook. I realize that is what the default filter is for, but they might have need of those groups on their own. I do not know how often we would run the lower level playbooks by themselves. The group_by module says it has the bypass_host_loop attribute, which, and I quote:

'Forces a ‘global’ task that does not execute per host, this bypasses per host templating and serial, throttle and other loop considerations

Conditionals will work as if run_once is being used, variables used will be from the first available host

This action will not work normally outside of lockstep strategies'

Does this attribute already give us the run_once quality that you were trying to achieve?

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+See Ansible's [playbooks best practices](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#handling-os-and-distro-differences) guide for more information on `group_by`
 
-CFS automatically sets this variable in the `hosts/01-cfs-generated.yaml` file for all sessions. When the session target is image customization, it sets `cray_cfs_image` to `true`; otherwise, it is `false`.
+### Using `cray_cfs_image` with conditionals
+
+When `group_by` cannot be used, `cray_cfs_image` can be set with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
+to selectively run individual tasks with `when: cray_cfs_image`, or to ignore individual tasks with `when: not cray_cfs_image`.
+
+Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#applying-when-to-roles-imports-and-includes)\).
```
jsollom-hpe, line 72: Your anchor on the URL is outdated. This URL brought me to the top of the page, same as if I had left off the #applying-when-to-roles-imports-and-includes.

They have two separate ones:
* (https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#conditionals-with-imports)
* https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#conditionals-with-includes

haasken-hpe (commented):
I think we could still be more clear about how exactly one should split up a playbook between node personalization and image customization.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+In instances where the same playbook may be run in both modes, it is best practice to include a conditional on all parts of the playbook. This is best done by placing the conditional on an `include_*` statement.
```
jsollom-hpe, line 73: With this line, I can't tell whether you're promoting the group_by module or the use of include_*. You mention group_by first and include_* second. You do push group_by in the second document.

jsollom-hpe (changes requested):

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
 For situations when both image customization and node personalization need to be part of the same playbook, CFS provides the `cray_cfs_image` variable to distinguish between
 node personalization and image customization. When this variable is set to `true`, it indicates that the CFS session is an image customization and the playbook is targeting an image.
 
-## Using `cray_cfs_image`
+It is best practice to include a default in Ansible roles for playbook and role portability because CFS injects this variable at runtime. This can be done in the defaults section of the role, or where the variable is called. For example:
 
-`cray_cfs_image` can be set with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
-to selectively run individual tasks with `when: cray_cfs_image`, or to ignore individual tasks with `when: not cray_cfs_image`.
+```text
+{{ cray_cfs_image | default(false) }}
```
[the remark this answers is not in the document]
rbak-hpe: I don't know for sure, but I suspect this isn't a problem because it's stock go templating.  I don't expect we're falling back on this though.  We inject this variable into the environment, so this is more of a safety/good practice thing.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
 See [Configuration Management Use Cases](Configuration_Management.md#use-cases) for more information about image customization and when it should be used.
 
-Ideally image customization playbooks should be separate from node personalization playbooks. This reduces the likelihood that Ansible content will run in both modes, and reduces the need for conditional checks in the playbooks.
+## Using `cray_cfs_image`
 
 For situations when both image customization and node personalization need to be part of the same playbook, CFS provides the `cray_cfs_image` variable to distinguish between
 node personalization and image customization. When this variable is set to `true`, it indicates that the CFS session is an image customization and the playbook is targeting an image.
 
-## Using `cray_cfs_image`
+It is best practice to include a default in Ansible roles for playbook and role portability because CFS injects this variable at runtime. This can be done in the defaults section of the role, or where the variable is called. For example:
```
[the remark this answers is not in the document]
rbak-hpe: Yes and no.  The point here is that there's more than one way to provide a default value.  A Jinja2 default is one way.  The other way is that you can provide a defaults file in the role that the task is called in.  What would have made this clearer for you?

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+  tasks:
+    - name: Classify hosts by whether they're an image
+      group_by:
+        key: image_{{ cray_cfs_image | default(false) | lower }}
+
+# Image customization
+- name: Image customization
+  hosts: all:&image_true
+  gather_facts: false
+  tasks:
```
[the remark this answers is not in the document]
rbak-hpe: Sure, I can make that switch.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+- name: group_by cray_cfs_image
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Classify hosts by whether they're an image
+      group_by:
+        key: image_{{ cray_cfs_image | default(false) | lower }}
+
+- import_playbook: playbook1.yaml
+- import_playbook: playbook2.yaml
```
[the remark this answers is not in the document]
rbak-hpe: Despite the initial plan, there's not actually a need to have separate playbooks for customization and personalization.  When we had that conversation I was forgetting that you can't dynamically import playbooks. What matters now, and how we get around the issue with conditionals, is using "group_by" to create host groups for image_true and image_false, and using those instead of conditionals.  I could show the internals of playbook1 and playbook2, but it doesn't matter if they separate customization/personalization or if they each have some of both, as long as they are using "hosts" and not "when".

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
rbak-hpe: Sure, I'll add something for this.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,24 +3,40 @@
...
+
+`** NOTE **` This option is no longer recommended and should only be used in small playbooks and one-off cases
+ as it is more efficient for Ansible to determine this at the host level rather than checking the `cray_cfs_image` variable for multiple tasks.
+
+CFS also provides the `cray_cfs_image` variable to distinguish between node personalization and image customization. 
+`cray_cfs_image` can be used with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
 to selectively run individual tasks with `when: cray_cfs_image`, or to ignore individual tasks with `when: not cray_cfs_image`.
 
-Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#applying-when-to-roles-imports-and-includes)\).
+Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#conditionals-with-includes).
```
jsollom-hpe, line 32: Your text says roles so use https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#conditionals-with-roles
instead of -includes.

Unless you explicitly meant the includes anchor because it dovetails with the next sentence.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,24 +3,40 @@
 The Configuration Framework Service \(CFS\) enables Ansible playbooks to run against both running nodes \(node personalization\) and images prior to boot\(image customization\).
 See [Configuration Management Use Cases](Configuration_Management.md#use-cases) for more information about image customization and when it should be used.
```
jsollom-hpe, line 4: In Configuration_Management.md this line should be
OLD: 'This use case enables partitioning a full configuration of a target node.'
NEW: 'This use case enables provisioning a full configuration of a target node.'
I just don't think partitioning is the verb you want.

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,67 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
 For more information on `gather_subset`, see the external [Ansible module setup](https://docs.ansible.com/ansible/latest/modules/setup_module.html) documentation.
-* Avoid importing playbooks in other playbooks.  This will trigger fact gathering for each imported playbook, potentially collecting the same information multiple times.
+* Reducing fact gathering time is especially important when importing multiple playbooks from a top level playbook.
+  Fact gathering will trigger for each imported playbook, potentially collecting the same information multiple times.
+  
+### `group_by` and `add_host`
+
+The `group_by` and `add_host` modules can both be used to dynamically generate new hosts groups for the Ansible inventory.  This is useful when for hosts that be grouped according to a common property, and when plays or large sections of plays are intended to only target a particular kind of host.  Examples of this might include grouping by operating system, hardware type, or a hardware property such as the presence of a GPU.
```
jsollom-hpe, line 54: Second sentence proposed replacement:
These modules proved useful when hosts can be group according to a common property. Then, plays can be designed to only target that particular group.

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,67 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+
+`group_by` should by used when there are multiple named groups that hosts can be grouped by.  For more instance on using this module see See Ansible's [playbooks best practices](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#handling-os-and-distro-differences) guide for more information.
```
jsollom-hpe, line 56: instance --> instances or possibly examples

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,67 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+
```
jsollom-hpe, line 57: Do you want to preface this with 'Here is an example of using the group_by module.'?

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,67 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+```
+
+`add_host` is useful for cases where the property is true or false.  It allows users to create a new group consisting of only the hosts where the property is true.  See Ansible's documentation on [](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/add_host_module.html) for more information.
+
+```yaml
+- name: group by a sample variable
+  hosts: all
+  tasks:
+    - name: Add all where sample_var is true to the new Sample group
```
jsollom-hpe, line 79: Add all hosts

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,67 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+        groups: sample_group
+      when: sample_var
+
+- name: Sample playbook
+  hosts: sample_group
+  roles:
+    ...
+```
+
+To target only a new group, plays should use the following syntax.  In this example the play is targeting only images for `Compute` nodes. `&` takes the intersection of the `Compute` and `sample_group` groups.
```
jsollom-hpe, line 91: I feel like this sentence is a little misleading because your earlier example shows targeting only a new group, the sample_group. The examples that follows targets the intersection of two groups.
Again, I would italicize 'images' to highlight it.

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,67 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+
+```yaml
+hosts: Compute:&sample_group
+```
+
+To target everything except the new group, plays should use the following syntax.  In this example the play is targeting only running `Compute` nodes. `!` negates the `sample_group` group, so that only Compute nodes that are not an image are targeted.
```
jsollom-hpe, line 97: Italicize 'running'

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -71,6 +129,16 @@ Some Ansible modules will optimize the command, such as grouping package install
...
 
+### Summary of the key tips
+
+For users just starting to write plays, or who just want to focus on the biggest improvements, here is a summary of the key tips:
+
+* Disable fact gathering if the play doesn't use Ansible facts: `gather_facts: false`
+* Use image customization where possible to avoid running tasks every time nodes boot.
+* Use the `cfs_image` host group to specify whether a play is intended for image customization or node personalization.
+* Use `group_by`, `add_host` and `include_*` to avoid repeating conditionals.
+* Use Ansible existing Ansible modules rather than calling shell commands or scripts.
```
jsollom-hpe, line 140: Delete first Ansible.

jsollom-hpe (changes requested):

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
rbak-hpe: Outdated

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
rbak-hpe: Outdated

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
rbak-hpe: I chose to point specifically to "includes" because that's the preferred method.  If you disagree with that those I can change it.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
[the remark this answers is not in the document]
jsollom-hpe: Well, if you were talking about just a Jinja 2 filter, I would have explicitly stated that.

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,74 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+        groups: sample_group
+      when: sample_var
+
+- name: Sample playbook
+  hosts: sample_group
+  roles:
+    ...
+```
+
+To target only a subset of a set of nodes, plays should use the following syntax.  In this example the play is targeting only _images_ for `Compute` nodes. `&` takes the intersection of the `Compute` and `sample_group` groups.
```
jsollom-hpe, line 98: Something went awry in the editing. You are using sample_group for the group when I think you want to use cfs_image. You do this in both image customization and node personalization examples. The text description references _images_ and _running_ which doesn't correlate with a random sample_group group.

jsollom-hpe (changes requested):

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,29 +3,75 @@
...
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Classify hosts by whether they're an image
+      group_by:
+        key: image_{{ cray_cfs_image | default(false) | lower }}
+
+# Image customization
+- name: Image customization
+  hosts: all:&image_true
```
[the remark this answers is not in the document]
haasken-hpe: Oh, I get it now. I didn't realize this was the Ansible syntax for an intersection of host groups as documented here: https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html#common-patterns

FWIW, I do think `hosts: Compute:&image_true` is a more realistic example for most.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,24 +3,41 @@
...
-Ideally image customization playbooks should be separate from node personalization playbooks. This reduces the likelihood that Ansible content will run in both modes, and reduces the need for conditional checks in the playbooks.
+## Using the `cfs_image` host group
 
-For situations when both image customization and node personalization need to be part of the same playbook, CFS provides the `cray_cfs_image` variable to distinguish between
-node personalization and image customization. When this variable is set to `true`, it indicates that the CFS session is an image customization and the playbook is targeting an image.
+During image customization, CFS will automatically add all image customization hosts into a special `cfs_image` host group in Ansible inventory.
+Plays intended for image customization can then target this group in addition to any other provided host groups.
 
-## Using `cray_cfs_image`
+To target only image customization, plays should use the following syntax.  In this example the play is targeting only _images_ for `Compute` nodes. `&` takes the intersection of the `Compute` and `cfs_image` groups.
```
haasken-hpe, line 11: Perhaps linking to this part of the Ansible docs is helpful here? https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html#common-patterns

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,24 +3,41 @@
...
 
-`cray_cfs_image` can be set with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
+```yaml
+hosts: Compute:&cfs_image
```
haasken-hpe, line 14: This is a little confusing to see outside the context of a play. E.g., lots of plays don't specify a name and so this would actually have a `- ` before it. Or if they do specify the name of the play first, then this would be indented by two spaces. E.g.:

```yaml
- hosts: Compute:&cfs_image
  roles:
  - sat-ncn
```

Or if there's a name:

```yaml
- name: Configure SAT on management nodes
  hosts: Compute:&cfs_image
  roles:
  - sat-ncn
```

Maybe a more complete play example would be beneficial? Anyway, pretty minor comment.

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,74 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
 * Turn off facts that are not needed in a playbook by setting `gather_facts: false`. If only a few facts are required, it is also possible to limit fact gathering by setting `gather_subset`.
 For more information on `gather_subset`, see the external [Ansible module setup](https://docs.ansible.com/ansible/latest/modules/setup_module.html) documentation.
-* Avoid importing playbooks in other playbooks.  This will trigger fact gathering for each imported playbook, potentially collecting the same information multiple times.
+* Reducing fact gathering time is especially important when importing multiple playbooks from a top level playbook.
+  Fact gathering will trigger for each imported playbook, potentially collecting the same information multiple times.
+  
+### `group_by` and `add_host`
+
+The `group_by` and `add_host` modules can both be used to dynamically generate new hosts groups for the Ansible inventory.
+These modules prove when hosts can be group according to a common property. Then, plays can be designed to only target that particular group.
```
haasken-hpe, line 55: typos here?

```suggestion
These modules prove useful when hosts can be grouped according to a common property. Then, plays can be designed to only target that particular group.
```

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,74 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+Examples of this might include grouping by operating system, hardware type, or a hardware property such as the presence of a GPU.
+
+`group_by` should by used when there are multiple named groups that hosts can be grouped by.
```
haasken-hpe, line 58: typo and dangling preposition that can be easily fixed

```suggestion
`group_by` should be used when there are multiple named groups by which hosts can be grouped.
```

operations/configuration_management/Write_Ansible_Code_for_CFS.md
```diff
@@ -38,15 +38,74 @@ CFS will handle scaling up Ansible to run on many hosts, but there are still pla
...
+        key: os_{{ ansible_facts['distribution'] }}
+
+- name: centOS playbook
+  hosts: os_CentOS
+  roles:
+    ...
+
+```
+
+`add_host` is useful for cases where the property is true or false.  It allows users to create a new group consisting of only the hosts where the property is true.  See Ansible's documentation on [](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/add_host_module.html) for more information.
```
haasken-hpe, line 78: empty brackets in this link

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
```diff
@@ -3,24 +3,41 @@
...
+ as it is more efficient for Ansible to determine this at the host level rather than checking the `cray_cfs_image` variable for multiple tasks. 
+ The preferred method is to use the aforementioned `cfs_image` host group.
+
+CFS also provides the `cray_cfs_image` variable to distinguish between node personalization and image customization. 
+`cray_cfs_image` can be used with [Ansible playbook conditionals](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html)
 to selectively run individual tasks with `when: cray_cfs_image`, or to ignore individual tasks with `when: not cray_cfs_image`.
 
-Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#applying-when-to-roles-imports-and-includes)\).
+Conditionals can also be applied to entire roles if desired \(see the external [apply Ansible conditionals to roles](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#conditionals-with-includes).
```
haasken-hpe, line 33: IMO, this could use a more complete example showing how to split up a playbook into one play for cfs_image and one for not cfs_image. Something like:

```yaml
- hosts: Management_Worker:&cfs_image
  any_errors_fatal: true
  remote_user: root
  roles:
    - cos-services-install

- hosts: Management_Worker:!cfs_image
  any_errors_fatal: true
  remote_user: root
  roles: 
    - cos-services-restart
```

And should this actually be using `include_role` in a task rather than a list of `roles`? If that's the recommendation, I definitely think an example would be helpful (obviously different from mine above which does static role import).

haasken-hpe (approved):
Approving with a few comments. I think this is more clear to me now. I'd like to see a couple more complete examples, and I'd like slightly more clarity on whether it's recommended to use `include_role` in a tasks list rather than a `roles` list. Other than that and Jason's comment about an editing or copy-paste issue, this looks good, so I'm approving.

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
rbak-hpe: I've added an example and clarified the use of `include_role`

operations/configuration_management/Target_Ansible_Tasks_for_Image_Customization.md
rbak-hpe: I've added an example, so I'm going to leave this as is.  This is really just for syntax and wasn't meant to show the playbook structure.

jsollom-hpe (approved):

PR #2555, opened 2022-09-27
CASMPET-6004: Add validation for checking the user input against etcd

mitchty:
This adds the ability for encryption.sh --enable/disable --aesgcm ... to validate the user input makes logical sense against what is found in the etcd database.

What this means is essentially if a user tries to setup encryption with a key that isn't found in etcd, the script will reject the update as being logically invalid.

This is to ensure that kubernetes secrets can always be read given a specific encryptionn configuration.

Also added switches --status to display the status of encryption.

And added a --restart switch to make restarting the daemonset easier to force updates.

# Description

Adds a layer of validation to prevent a user from setting up an encryption configuration that would be invalid or result in existing encrypted secrets from being readable by the kube api server.

# Checklist Before Merging

### DIFF
--- a/operations/kubernetes/encryption/README.md
+++ b/operations/kubernetes/encryption/README.md
@@ -6,6 +6,8 @@ This controller is deployed by default in CSM with the `cray-kubernetes-encrypti
 
 By default, encryption is not enabled and must be enabled after install.
 
+Note that control plane is used in this document elsewhere master or management nodes may be used. Control plane is used to be consistent with upstream Kubernetes documentation.
+
 ## Table of contents
 
 * [Implementation details](#implementation-details)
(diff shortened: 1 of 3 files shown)


dzou-hpe (approved):

mitchty:
Found a race condition if a secret is written when removing encryptino but not all nodes have been updated yet, will update the pr with a fix for this condition.

```
ncn-m003:~ # encryption.sh --disable --aescbc sutsutsutsutsutssutsutsutsutss
encryption.sh                                                                                                100%   24KB  14.0MB/s   00:00
fatal: requested goal conflicts with existing etcd encryption
etcd: aescbc-be59cdb38ed61f468deecaf865643be643201906504dc8f3e85e432d5463d525 identity
requested: identity aescbc-be59cdb38ed61f468deecaf865643be643201906504dc8f3e85e432d5463d525
To fix add the keys that correspond with that encryption secret to the end of this command. Failure to do so would mean k8s cannot read existing secrets.
```

This should be allowed but my logic is a bit perhaps too strict in my set operations.

rgross727 (approved):

rgross727 (approved):

rustydb (approved):

PR #2556, opened 2022-09-27
CASMINST-5395: Fix bug in prerequisites.sh; Retry curl command; include stage names in upgrade log; remove unused script

mharding-hpe:
# Description

When the [CASMINST-5347](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5347) PRs were reverted, it wasn't a targeted revert of the changes which were causing the problems, but just a blanket revert of all of it. That included reverting some linting and other fixes. This PR restores those portions. Specifically:

- Remove the worker rebuild script which is no longer used
- Include the upgrade stage names in the upgrade log file, to make it easier to trace execution from the log
- Move the Bash function sourcing in prepare-assets to the same location, after necessary variables have been set
- Retry a curl call whose transient failure caused the upgrade to fail on surtur
- Remove the whitespace from the "UPDATE NCN KERNEL PARAMETERS" stage name, to match all other stage names. This is also currently causing the prerequisites script to behave incorrectly with regards to that step. Specifically, that step will always be re-run if prerequisites.sh is re-run on ncn-m001. That is because of a second issue (which this PR does not correct, in the interest of limiting the scope of the changes) -- namely that record_state is called without quoting the ${stage_name} variable (and the record_state function does not validate that it received the expected number of arguments). Without this PR in place, the stage name has whitespace in it, so the record_state function thinks it is being called for a stage named UPDATE on a host named NCN. But because the function used to check the states is properly quoting its argument variables, it looks for the full stage name on ncn-m001, which it will never find (because of the just-disclosed issue). This PR does not correct all of the issues listed above, but it corrects enough of them to avoid the problems described above.

None of the above changes were related to the problems that caused CASMINST-5347 to be reverted.

# Checklist Before Merging


rustydb (approved):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -188,13 +188,13 @@ if [[ $state_recorded == "0" && $(hostname) == "ncn-m001" ]]; then
     record_state ${state_name} "$(hostname)"
```
dzou-hpe, line 188: i'd `tee` these ones to the log file as well cause this is what tells you a stage has been done. if we only record the start, we would need to guess whether it has been done or now

dzou-hpe (approved):

upgrade/scripts/upgrade/prerequisites.sh
mharding-hpe: Good idea!

mharding-hpe: Done. Let me know how it looks to you now.

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -184,17 +183,17 @@ if [[ $state_recorded == "0" && $(hostname) == "ncn-m001" ]]; then
     echo "Taking snapshot of current Kubernetes pod states to ${K8S_PODS_SNAPSHOT}"
     kubectl get pods -A -o wide --show-labels > "${K8S_PODS_SNAPSHOT}"
 
-    } >> ${LOG_FILE} 2>&1
-    record_state ${state_name} "$(hostname)"
+    } >> "${LOG_FILE}" 2>&1
+    record_state "${state_name}" "$(hostname)" | tee -a "${LOG_FILE}"
```
leliasen-hpe, line 187: I'm just curious if there is a reason that `record_state "${state_name}" "$(hostname)"` is outputted to the LOG_FILE in this PR to release/1.3 and not in the PR to main?

leliasen-hpe (approved):

PR #2559, opened 2022-09-27
Removed prompt from bash commands

mantoanibroadwing:
# Description

This is to remove `pit#` and `ncn#`  prompt from commands presented on documentation, with this will be easier to copy commands with multiple lines during a new install or upgrade.

# Checklist Before Merging

- [X ] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

### DIFF
--- a/install/bootstrap_livecd_remote_iso.md
+++ b/install/bootstrap_livecd_remote_iso.md
@@ -106,8 +106,8 @@ On first login (over SSH or at local console) the LiveCD will prompt the adminis
 1. Start a typescript to record this section of activities done on `ncn-m001` while booted from the LiveCD.
 
    ```bash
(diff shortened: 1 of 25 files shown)


rustydb (changes requested):
If we want to do this for 1.2, we still need to keep the context like we do in 1.3/1.4 documentation.

For example see this page, notice the (`pit#`) items https://github.com/Cray-HPE/docs-csm/blob/main/introduction/documentation_conventions.md#command-prompt-location

I am a huge fan of doing this in 1.2, but we'll just want it to match how we do things in 1.3 and 1.4.

rustydb:
@mantoanibroadwing I'll usher this through before end of November, might seem late but personally I want this before I do 1.2 installs later.

PR #2596, opened 2022-10-04
CASMTRIAGE-4277 Logging improvements for Argo workflows and a few commands.

mitchty:
# Description

First two words. DON'T PANIC This appears like a lot of code, and it is, but a lot of it is unit tests and other ancillary things to enable the following explanation. Note the lib.sh.yaml files are just redundant code. Ignore them entirely.

This is a huge amount of change but in the end it amounts to the following:

Removing some (not all there are like 120+ usages in docs-csm, not enough time to fix them all) curl | jq pipes for certain setups in argo and outside of it. 

And replaces them with basically:

- make a temp file
- curl -o to that tempfile
- jq from that file 

That is it, all the rest is scaffolding and code to allow for a wrapper curl and jq trampoline functions in shell to then take over doing the job of calling curl and validating that curl didn't (by default):
- return non 0
- return without an http 200 response upstream

Similarly for jq it will now run jq on the tempfile first to establish that the file is indeed, json. Iff not it exits.

What this does is allows us to setup hook functions that can control what and how a usage of curl in shell exits.

So if we had a call like curl http://example.com and for some reason, lets say dns doesn't resolve. We would end up with the script exiting at that curl() function call by spitting out to stderr something like this:
command:
curl http://example.com
rc:
?whatever curl would return in this instance
output:
000

Similarly if the jq trampoline function is fed a file which is curl()'d from ^^^ and finds that the file is... not json lets say, it will exit and dump a similar message to stderr.

command:
jq . /filename
rc: 4
output: parse error: Invalid numeric literal at line 1, column 2
file /notjson content:
I am not json

# Checklist Before Merging


mitchty:
Note, I just realized I need to make sure these jq/curl trampoline functions don't impact existing jq/curl users that don't abide by these new rules. I'll add tests for that as well and validate on fanta as I think some of these other argo calls might use that. Hence why this is a draft for now.

mitchty:
This is going to be done in a different way. Closing

PR #2740, opened 2022-10-27
CASMPET-6010 Add IUF support for Nexus setup

rnoska-hpe:
# Description

This adds a new Argo workflow template to support the Install and Upgrade Framework (IUF) ability to create Nexus blob stores and repositories as required.

# Checklist Before Merging

### DIFF
--- a/workflows/iuf/nexus-setup/nexus-setup-template.yaml
+++ b/workflows/iuf/nexus-setup/nexus-setup-template.yaml
@@ -34,11 +34,90 @@ spec:
     # The value for content is set to the json string value for 'content' in the manifest when --parameter-file is
     # set to the path of the IUF manifest yaml file.
     - name: content
-  entrypoint: nexus-setup
+  entrypoint: main
   templates:
-  - name: nexus-setup
+### Main Steps ###
+  - name: main
+    steps:
+    - - name: nexus-get-admin-credential
+        template: nexus-get-admin-credential-template
+    - - name: nexus-setup
+        template: nexus-setup-template
+        arguments:
+          parameters:
+          - name: nexus_admin_credential_secret_name
+            value: "{{steps.nexus-get-admin-credential.outputs.parameters.secret_name}}"
+          - name: content
+            value: "{{workflow.parameters.content}}"
+          - name: product
+            value: "{{workflow.parameters.product}}"
+          - name: product_host_path
+            value: "{{workflow.parameters.product_host_path}}"
+    - - name: cleanup
+        template: cleanup-template
+        arguments:
+          parameters:
+          - name: nexus_admin_credential_secret_name
+            value: "{{steps.nexus-get-admin-credential.outputs.parameters.secret_name}}"
+### Templates ###
+## nexus-get-secret-template ##
+  - name: nexus-get-admin-credential-template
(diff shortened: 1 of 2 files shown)


workflows/iuf/nexus-setup/nexus-setup-template.yaml
```diff
@@ -0,0 +1,184 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: nexus-setup-template
+  namespace: argo
+spec:
+  arguments:
+    parameters:
+    - name: nexus-setup-image
+    - name: product
```
dzou-hpe, line 33: @atifsyedali i think we said the input will be just one big json string, right? the operator won't really know this detailed input requirements

workflows/iuf/nexus-setup/nexus-setup-template.yaml
```diff
@@ -0,0 +1,184 @@
...
+    script:
+      image: artifactory.algol60.net/csm-docker/stable/docker.io/portainer/kubectl-shell:latest-v1.21.1-amd64
+      command: [bash]
+      source: |
+        function sync_item() {
+          item_name="$1"
+          source_ns="$2"
+          destination_name="$3-$RANDOM"
+          destination_ns="$4"
+          if kubectl get $item_name -n $source_ns &> /dev/null; then
```
dzou-hpe, line 97: @rnoska-hpe are we sure this pod will have enough permission to access secret in other ns?

workflows/iuf/nexus-setup/nexus-setup-template.yaml
```diff
@@ -0,0 +1,184 @@
...
+kind: WorkflowTemplate
+metadata:
+  name: nexus-setup-template
+  namespace: argo
+spec:
+  arguments:
+    parameters:
+    - name: nexus-setup-image
+    - name: product
+    - name: product_host_path
```
dzou-hpe, line 34: i think this will be `web_dir` +`product` web_dir is a global parameter we will always pass (in the big json string)

dzou-hpe (approved):

workflows/iuf/nexus-setup/nexus-setup-template.yaml
rnoska-hpe: Yes, this was tested and worked using Argo as it was installed on Mug and Leomdrop (csm-1.3). This is also essentially the same code that is used in sontar to copy items across namespaces.

workflows/iuf/nexus-setup/nexus-setup-template.yaml
rnoska-hpe: So that I can close on this and address this feedback, I think it would be best to create another Jira to integrate this with the operator when that is available.

workflows/iuf/nexus-setup/nexus-setup-template.yaml
rnoska-hpe: I think it would be best to create another Jira to integrate this with the operator when that is available (same as above).

atifsyedali (approved):

rustydb (approved):

PR #2757, opened 2022-10-31
Revert "CASMPET-6111: Update arg parsing of this script to be more strict"

denniswalker:
Reverts Cray-HPE/docs-csm#2750


rustydb (approved):

denniswalker (commented):
Reverting the change until we see it work well in a 1.4 upgrade.

PR #2779, opened 2022-11-03
CASMINST-5480: Correct problems with NCN personalization procedure for standard upgrade case

mharding-hpe:
# Description

The current NCN personalization documentation does not make sure that the NCNs use the new CFS configuration as their desired configuration. In addition to overall linting and reorganization of that procedure, this corrects that oversight.

# Required Reviewers

## Don Bahls / Dean Roe: 

I also noticed that as written, the current procedure neither enabled nor disabled the NCNs in CFS. In the case of CSM-only upgrades, we disable them in CFS, and they are then configured after they are rebuilt. I modified this NCN personalization procedure to work the same way (when being done as part of a CSM upgrade). However, I do not know if any of the non-CSM products have dependencies that mean they should run CFS during the CSM upgrade before being rebuilt. If that is the case, then I assume it is only the case for worker nodes, and could modify the procedure to enable CFS on the worker nodes before they are rebuilt. I'll need someone like Dean Roe or Don Bahls to weigh in on that.

## SAT

Since this is a re-write of a procedure centered around `sat bootprep`, someone from that team should make sure I haven't done anything dumb without realizing it.

## SSI Upgraders

Ideally Jeff Sparrow could review this, since he had a hell of a time with it on hermod. But anyone who does CSM upgrades with other products present would be great reviewers. I want to know if this is any clearer than what we had before.

# Checklist Before Merging


rustydb:
Will wait for SAT SME.

mharding-hpe:
> Will wait for SAT SME.

Also need to wait for Dean Roe or @don-bahls-hpe to look at it and answer my question for them in the PR description

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -1,25 +1,60 @@
...
+    If a CFS configuration with this name already exists, the procedure on this page will overwrite it.
+
+    Because this procedure will create/update a CFS configuration and modify the CFS components of the management NCNs,
+    take a snapshot of the current state of these CFS objects.
+
+    ```bash
+    /usr/share/doc/csm/scripts/operations/configuration/backup_cfs_config_comp.sh
+    ```
 
 1. (`ncn-m#`) Perform the steps in the [Accessing `sat bootprep` Files](Accessing_Sat_Bootprep_Files.md) procedure to gather a copy of the `sat bootprep` files.
```
mharding-hpe, line 46:
```suggestion
1. (`ncn-m#`) Gather a copy of the `sat bootprep` files.

    See [Accessing `sat bootprep` Files](Accessing_Sat_Bootprep_Files.md).
```

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -1,25 +1,62 @@
...
-The following procedure describes how to correctly edit the `sat bootprep` files to be able to use them to perform node personalization.
+## Background
 
-1. (`ncn-m#`) Perform the steps in the [Accessing `sat bootprep` Files](Accessing_Sat_Bootprep_Files.md) procedure to gather a copy of the `sat bootprep` files.
+NCN node personalization refers to the process of CFS applying a configuration to a management NCN after it is booted.
+During CSM installs and upgrades, this CFS configuration must be created and set as the desired configuration in CFS for all management NCNs.
+The purpose is to ensure that the appropriate CFS layers are applied to the management NCNs after they boot.
 
-1. (`ncn-m#`) Create a local copy of the `management-bootprep.yaml` file and delete the `ncn-image-customization` configuration. The `ncn-personalization` configuration should be the only entry remaining in the file if completed correctly.
+The same CFS configuration is used for post-boot personalization of master, storage, and worker NCNs. However, some individual parts of that
```
don-bahls-hpe, line 18: is there a reason for this line break between line 18 and 19?

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -29,62 +66,151 @@ The following procedure describes how to correctly edit the `sat bootprep` files
...
+    - name: cpe-pe_deploy-integration-{{cpe.version}}
+      playbook: pe_deploy.yml
+      git:
+        url: https://api-gw-service-nmn.local/vcs/cray/cpe-config-management.git
+        commit: c38c16bc9a645a9339a8303865e696b089b15e17
+    - name: analytics-site-integration-{{analytics.version}}
+      playbook: site.yml
+      git:
+        url: https://api-gw-service-nmn.local/vcs/cray/analytics-config-management.git
+        commit: c9c0b2cc69998830a47d7c989b07f550814af095
```
don-bahls-hpe, line 154: Do we want to remove the `{{cpe.version}}` and `{{analytics.version}}` above?   If we leave this as is, the name will be the papaya version but the commit id will be the raspberry version.

scripts/operations/configuration/backup_cfs_config_comp.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/bash
+#
```
don-bahls-hpe, line 2: Maybe I'm missing something but this seems like a lot of code to have to maintain just to reduce the number of command run from two to one.   Isn't this essentially just running "cray cfs configurations list" and "cray cfs components list"?

operations/configuration_management/NCN_Node_Personalization.md
mharding-hpe: Line length. Like HTML, when the markdown is rendered, the line break goes away. So our source files usually start new lines somewhere after 60-80-ish characters.

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -29,62 +66,151 @@ The following procedure describes how to correctly edit the `sat bootprep` files
     - name: ncn-personalization
     ```
 
-1. In the case of an upgrade, execute the following two substeps:
+### 2. Edit CPE and Analytics layers
+
+This section is required if this procedure is being performed as part of a CSM upgrade.
```
ekamin-hpe, line 71: I feel like this section wouldn't be required if this procedure was being performed as part of a CSM*-only* upgrade, I think when I wrote "in the case of an upgrade" I was thinking more like a full recipe upgrade.

operations/configuration_management/NCN_Node_Personalization.md
mharding-hpe: Good call.

scripts/operations/configuration/backup_cfs_config_comp.sh
mharding-hpe: Yeah. But we've had issues before where if we just include commands to redirect output to a file, we need to make sure that we aren't overwriting existing files, and are putting them in sensible locations. Similarly, there have been commands which failed in non-obvious ways, and most folks don't check return codes at the command line. Those are the two main reasons I favor automating even this kind of thing.

The other reason is that most of the functions defined in here are actually going to be in a common shell function library, but for 1.3 I didn't want to do that in the interest of limiting the size of the PR.

mharding-hpe: I appreciate your attention to detail in your review, but my two biggest things I want your eyes for are:

1. Is it okay for me to leave the NCNs disabled in CFS until after they rebuild? For CSM it's fine, but do any of the other products need the updated NCN personalization to be run on the NCNs BEFORE they are rebuilt? 

2. Is this new procedure easier to follow?

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -29,62 +66,151 @@ The following procedure describes how to correctly edit the `sat bootprep` files
...
+If that is not the case, then skip ahead to [Update CFS configuration and components](#4-update-cfs-configuration-and-components).
+
+In this section, the `management-bootprep-node-personalization.yaml` file is modified to specify the versions of CPE and Analytics that
+are currently in use on the system. This is necessary because the new versions of CPE and Analytics have not yet been installed at this
+time in the CSM upgrade procedure.
+
+1. (`ncn-m#`) Acquire a copy of the current CPE and Analytics products CFS configuration values already in use.
+
+    There are multiple ways to do this. The following is an example of finding these values by looking at the CFS configuration of a worker NCN on the system.
```
ekamin-hpe, line 80: Do you have any concern here about the potential for some NCNs to have different configurations?

Not saying it has to be done with SAT and as you say it can be done different ways, but using `sat status` here would tell you the desired configurations for all the NCNs at once.

```
ncn-m001:~ # sat status --filter role=management --fields xname,desiredconfig
+---------------+---------------------+
| xname         | Desired Config      |
+---------------+---------------------+
| x3000c0s1b0n0 | ncn-personalization |
| x3000c0s2b0n0 | ncn-personalization |
| x3000c0s3b0n0 | ncn-personalization |
| x3000c0s4b0n0 | ncn-personalization |
| x3000c0s5b0n0 | ncn-personalization |
| x3000c0s6b0n0 | ncn-personalization |
| x3000c0s7b0n0 | ncn-personalization |
| x3000c0s8b0n0 | ncn-personalization |
| x3000c0s9b0n0 | ncn-personalization |
+---------------+---------------------+
```

And if you wanted to be really fancy:
```
ncn-m001:~ # sat status --filter role=management --fields desiredconfig --format json | jq -r '.[]."Desired Config"' | uniq
ncn-personalization
```

Scrolling down in the procedure a little I see that it doesn't REALLY matter that much because the procedure will have you update the `desiredConfig` for all the nodes at once to this config anyway, so by the end they will all have the same config.

mharding-hpe:
@don-bahls-hpe I truly do appreciate your attention to detail in your review, but my two biggest things I want your eyes for are:

1. Is it okay for me to leave the NCNs disabled in CFS until after they rebuild? For CSM it's fine, but do any of the other products need the updated NCN personalization to be run on the NCNs BEFORE they are rebuilt? 

2. Is this new procedure easier to follow?

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -29,62 +66,151 @@ The following procedure describes how to correctly edit the `sat bootprep` files
...
+        ncn-personalization
+        ```
 
-    1. (`ncn-m#`) Acquire a copy of the current CPE and Analytics products CFS configuration values already in use.
-     Obtain the values for the `cloneUrl`, `commit`, and `playbook` lines for each of those two layers in the next step to ensure that they are personalized with the desired configuration for CPE and Analytics.
+    1. Show the CPE and Analytics layers of that configuration.
 
         ```bash
-        cray cfs components describe <ncn-xname> --format json
+        cray cfs configurations describe "${WORKER_CONFIG}" --format json | \
```
ekamin-hpe, line 111: I'm totally good with it and you are probably aware, but I just wanted to point out that this is different than what we had - we were looking at `cray cfs components describe` in order to see what configuration had last successfully run for the component, whereas this is looking at the desired config layers for the node.

If anything I think this might be better to look at a configuration rather than the state of an individual component.

I don't recall whether we talked any more about the pros and cons, the closest I can find is this slack thread: https://cray.slack.com/archives/CLG4R6SU9/p1665604546333739

ekamin-hpe (approved):
This looks good from a SAT perspective, all the bootprep stuff looks good.

I only had a few comments but I don't think they were huge issues.

operations/configuration_management/NCN_Node_Personalization.md
mharding-hpe: Yeah, I don't think we discussed it, but my line of thought when I came to that part was exactly like you laid out here. I figured that the configuration was a better source of truth. I had meant to ask you about that but forgot, so I'm glad you spotted it.

operations/configuration_management/NCN_Node_Personalization.md
mharding-hpe: Using sat for this definitely seems simpler, so I'll make that edit. And if they are using multiple configurations, then they'd probably want to know that so they could make sure that they are using the one they wanted as the source. Good call.

operations/configuration_management/NCN_Node_Personalization.md
mharding-hpe: For sure. The CSM upgrade docs make that clear -- this page is only linked from the section that is for what we call the "standard upgrade", which is the non-CSM-only case. That said, it wouldn't hurt for me to be clearer here as well, to reduce the possibility of confusion. Good catch.

mharding-hpe:
> This looks good from a SAT perspective, all the bootprep stuff looks good.
> 
> I only had a few comments but I don't think they were huge issues.

Thanks a lot for all of your comments. They are all good suggestions/observations and are exactly what I was hoping for. They'll definitely result in improvements. But I may not have the PR updated with those improvements until tomorrow.

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -1,25 +1,62 @@
...
+  1. [Cleanup](#5-cleanup)
 
-The definition of the CFS configuration used for node personalization is provided in the `hpc-csm-software-recipe` repository in VCS.
-The following procedure describes how to correctly edit the `sat bootprep` files to be able to use them to perform node personalization.
+## Background
 
-1. (`ncn-m#`) Perform the steps in the [Accessing `sat bootprep` Files](Accessing_Sat_Bootprep_Files.md) procedure to gather a copy of the `sat bootprep` files.
+NCN node personalization refers to the process of CFS applying a configuration to a management NCN after it is booted.
+During CSM installs and upgrades, this CFS configuration must be created and set as the desired configuration in CFS for all management NCNs.
+The purpose is to ensure that the appropriate CFS layers are applied to the management NCNs after they boot.
```
mharding-hpe, line 16:
```suggestion
NCN node personalization refers to the process of CFS applying a configuration to a management NCN after it is booted.
During CSM installs and upgrades[^1], this CFS configuration must be created and set as the desired configuration in CFS for all management NCNs.
The purpose is to ensure that the appropriate CFS layers are applied to the management NCNs after they boot.

[^1]: Except for CSM-only installs and upgrades, which are very rare in production environments. If CSM is the only product
on the system, then this procedure is not performed.
```

don-bahls-hpe:
> Also need to wait for Dean Roe or @don-bahls-hpe to look at it and answer my question for them in the PR description

@mharding-hpe sorry I missed where the ask was being called out, but I see it now.   
(1) For worker nodes I think it's fine to leave CFS disabled for the NCNs so long as it get re-enabled with the new ncn-personalization config as the rebuild of each worker node completes.    This should allow a DVS quorom to be maintained.   The only way that would fall apart would be if the system lost power during the upgrade, but that would likely cause all sorts of other issues.   

(2) What's here looked good to me, however I've not personally been actively involved in this part of an upgrade for papaya, so I can't comment on the relative ease differences vs previous instructions.

mharding-hpe:
I have implemented the suggestions from Don and Eli (although they both also indicated that the doc was fine even without the changes).

mharding-hpe:
Jeff Sparrow reviewed this and gave approval on Slack. I guess he doesn't have an account here so could not approve it directly.

This is ready for merge as far as I am concerned.

rustydb (approved):

nrockershousen (approved):

PR #2806, opened 2022-11-09
CASM-3632: Move some common script functions into shared library

mharding-hpe:
# Description

This PR has no functional changes. It just reorganizes some of the scripts used in NCN personalization and Vault updates. This is to enable future work.

I tested the modified scripts on drax and ensured that they still work as before.

It would be fine backporting this to 1.3, but since it does involve script changes, it is fine if we keep it to the main branch. We can always backport it later, if desired.

# Checklist Before Merging

### DIFF
--- a/scripts/operations/configuration/apply_csm_configuration.sh
+++ b/scripts/operations/configuration/apply_csm_configuration.sh
@@ -23,33 +23,10 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 
-err()
-{
-    echo "ERROR: $*" 1>&2
-}
-
-err_exit()
-{
-    err "$@"
-    exit 1
-}
-
-run_cmd()
-{
-    "$@" || err_exit "Command failed with rc $?: $*"
-}
-
-run_mktemp()
-{
-    tmpfile=$(mktemp "$@") || err_exit "Command failed with rc $?: mktemp $*"
-    [[ -n "${tmpfile}" ]] || err_exit "mktemp command passed but gave no output"
-    [[ -e "${tmpfile}" ]] || err_exit "mktemp command passed but '${tmpfile}' does not exist"
-    if [[ $# -gt 0 && "$1" == "-d" ]]; then
-        [[ -d "${tmpfile}" ]] || err_exit "mktemp -d command passed and '${tmpfile}' exists, but is not a directory"
-    else
-        [[ -f "${tmpfile}" ]] || err_exit "mktemp command passed and '${tmpfile}' exists, but is not a regular file"
-    fi
-}
+locOfScript=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+# Inform ShellCheck about the file we are sourcing
+# shellcheck source=./bash_lib/common.sh
+. "${locOfScript}/bash_lib/common.sh"
 
 usage()
 {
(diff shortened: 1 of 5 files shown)


scripts/operations/configuration/bash_lib/common.sh
```diff
@@ -0,0 +1,78 @@
...
+# Returns 0 if argument is a defined Bash function, 1 otherwise
+is_function()
+{
+    [[ $(LC_ALL=C type -t "$1") == function ]] && return 0 || return 1
+}
+
+# Print an error message to stderr
+err()
+{
+    echo "ERROR: $*" 1>&2
```
mitchty, line 37: >&2 to print just to stderr

mharding-hpe: Good call. Done!

scripts/operations/configuration/python_lib/__init__.py
```diff
@@ -0,0 +1,24 @@
+#
```
ekamin-hpe, line 1: Nothing wrong here, but I would point out that you could just leave this file empty unless you really want to document the module in the `__init__.py` doc string or something.

As of https://github.com/Cray-HPE/license-checker/commit/161ab1dbfba509d4dd89826f70ca21bf9fd7e0f8 the license checker passes empty files.

rustydb (approved):

scripts/operations/configuration/apply_csm_configuration.sh
```diff
@@ -23,33 +23,10 @@
...
-    tmpfile=$(mktemp "$@") || err_exit "Command failed with rc $?: mktemp $*"
-    [[ -n "${tmpfile}" ]] || err_exit "mktemp command passed but gave no output"
-    [[ -e "${tmpfile}" ]] || err_exit "mktemp command passed but '${tmpfile}' does not exist"
-    if [[ $# -gt 0 && "$1" == "-d" ]]; then
-        [[ -d "${tmpfile}" ]] || err_exit "mktemp -d command passed and '${tmpfile}' exists, but is not a directory"
-    else
-        [[ -f "${tmpfile}" ]] || err_exit "mktemp command passed and '${tmpfile}' exists, but is not a regular file"
-    fi
-}
+locOfScript=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
ekamin-hpe, line 26: I feel like we could probably debate this but personally I think using `cd` and `pwd` is sort of overkill in this case. Is it just to check that the directory exists?

ekamin-hpe (approved):
LGTM

scripts/operations/configuration/python_lib/__init__.py
mharding-hpe: Oh, cool -- I just copied this from another file in the repo. No need to put it there if it isn't needed

scripts/operations/configuration/apply_csm_configuration.sh
mharding-hpe: I copied this from another script in the docs repo, but I believe it is done to avoid getting a relative path to the file. I figured I'd stay consistent with how the other scripts were doing it, if for no other reason than to avoid having all of the scripts look a little different. I don't know if there is an edge case in there that they are trying to avoid by this mechanism.

ekamin-hpe: Good with me, didn't realize this was also how other scripts were doing it.

mharding-hpe: Yeah, the ones in upgrade/scripts all do it that way.

It looks to me like it was probably lifted from here: https://stackoverflow.com/a/246128

ekamin-hpe: Whoever wrote this should have read the article rather than just copying the top solution 😆

PR #2820, opened 2022-11-15
CASM-3653: write_root_secrets_to_vault: Add logging, split into modules

mharding-hpe:
# Description

This PR modifies the write_root_secrets_to_vault.py script.

From a user perspective, the script is unchanged (beyond some minor changes to the printed output). The changes are primarily splitting it into a few library modules, adding logging, and adding type hinting. This is to enable upcoming work as part of [CASM-3367](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3367).

I tested the updated script on fanta and verified that it still works as it did prior to the changes.

# Checklist Before Merging

### DIFF
--- a/scripts/operations/configuration/python_lib/api_requests.py
+++ b/scripts/operations/configuration/python_lib/api_requests.py
@@ -1,0 +1,101 @@
+#
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""Shared Python function library: API requests"""
+
+import logging
+import time
+import traceback
+from typing import Callable
+
+from . import common
+
+
+def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> None:
+    """
+    1) If a parent exception is passed in, make a debug log entry with its stack trace.
+    2) Log an error with the specified message.
+    3) Raise a ScriptException with the specified message (from the parent exception, if
+       specified)
+    """
+    if parent_exception is not None:
+        logging.debug(traceback.format_exc())
+    logging.error(msg)
+    if parent_exception is None:
+        raise common.ScriptException(msg)
+    raise common.ScriptException(msg) from parent_exception
+
+
+def make_api_request_with_retries(request_method: Callable,
+                                  url: str,
+                                  expected_status_code: int,
+                                  **request_kwargs):
+    """
+    Makes request with specified method to specified URL with specified keyword arguments (if any).
+    If the expected status code is returned, then the response body is returned (or None, if empty).
+    If a 5xx status code is returned, the request will be retried after a brief wait, a limited
+    number of times.
+    If the expected status code is never returned (or an unexpected and non-5xx status code is
+    ever returned), then raise an exception
+    """
+    try:
+        method_name = request_method.__name__.upper()
+    except Exception as e:
+        log_error_raise_exception(
+            "Unexpected error determining API request method name", e)
+
+    count = 0
+    max_attempts = 6
+    while count < max_attempts:
+        count += 1
+        logging.info(f"Making {method_name} request to {url}")
+        try:
+            resp = request_method(url, **request_kwargs)
+        except Exception as e:
+            log_error_raise_exception(
+                f"Error making {method_name} request to {url}", e)
+        logging.debug(f"Response status code = {resp.status_code}")
+        if resp.status_code == expected_status_code:
+            if resp.text:
+                try:
+                    return resp.json()
+                except Exception as e:
+                    logging.debug(f"Response reason = {resp.reason}")
+                    logging.debug(f"Response text = {resp.text}")
+                    log_error_raise_exception(
+                        f"Error parsing {url} {method_name} response as JSON", e)
+            else:
+                return None
+        logging.debug(f"Response reason = {resp.reason}")
+        logging.debug(f"Response text = {resp.text}")
+        if not 500 <= resp.status_code <= 599:
+            log_error_raise_exception(
+                "Unexpected status code received in response to API request."
+                f"Received {resp.status_code}, expecting {expected_status_code}")
+        elif count >= max_attempts:
+            log_error_raise_exception(
+                f"API request unsuccessful even after {max_attempts} attempts")
+        logging.info("Sleeping 3 seconds before retrying..")
+        time.sleep(3)
+    log_error_raise_exception(
+        "PROGRAMMING LOGIC ERROR: make_api_request_with_retries function should get here")
--- a/scripts/operations/configuration/write_root_secrets_to_vault.py
+++ b/scripts/operations/configuration/write_root_secrets_to_vault.py
@@ -23,120 +23,66 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 
-# Reads in:
-# - the hashed root user password from /etc/shadow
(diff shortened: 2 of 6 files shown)


scripts/operations/configuration/python_lib/api_requests.py
```diff
@@ -0,0 +1,84 @@
...
+                except Exception as e:
+                    logging.debug(f"Response reason = {resp.reason}")
+                    logging.debug(f"Response text = {resp.text}")
+                    log_error_raise_exception(f"Error parsing {url} {method_name} response as JSON", e)
+            else:
+                return None
+        logging.debug(f"Response reason = {resp.reason}")
+        logging.debug(f"Response text = {resp.text}")
+        if not 500 <= resp.status_code <= 599:
+            log_error_raise_exception("Unexpect status code received in response to API request")
```
ekamin-hpe, line 79:
```suggestion
            log_error_raise_exception("Unexpected status code received in response to API request")
```

scripts/operations/configuration/python_lib/api_requests.py
```diff
@@ -0,0 +1,84 @@
...
+        elif count >= max_attempts:
+            log_error_raise_exception(f"API request unsuccessful even after {max_attempts} attempts")
+        logging.info("Sleeping 3 seconds before retrying..")
+        time.sleep(3)
```
ekamin-hpe, line 83: `time` needs to be imported.

scripts/operations/configuration/write_root_secrets_to_vault.py
```diff
@@ -147,63 +143,20 @@ def main():
...
+        password_hash = get_root_hash_from_etc_shadow()
+        vault.write_root_secrets(ssh_private_key=ssh_private_key,
+                           ssh_public_key=ssh_public_key,
+                           password_hash=password_hash,
+                           verify=True)
+    except common.ScriptException as e:
+        common.print_err_exit(f"{e}")
+
+    print("SUCCESS", flush=True)
+    sys.exit(0)
```
ekamin-hpe, line 158: nit: this line is not necessary

scripts/operations/configuration/write_root_secrets_to_vault.py
```diff
@@ -38,105 +38,101 @@
...
-        err_exit("File '%s' only contains %d characters" % (filename, len(contents)))
-    return contents
-
-def get_root_hash_from_etc_shadow():
+import traceback
+
+from python_lib import common
+from python_lib import vault
+
+class _ExcludeErrorsFilter(logging.Filter):
```
ekamin-hpe, line 50: Just as an aside (doesn't have to be done now) the logging config might be nice to factor into another module (e.g. `logging.py` or something) in the future.

scripts/operations/configuration/python_lib/api_requests.py
ekamin-hpe: You also could log the value of `resp.status_code` in the error message, that could be helpful.

Overall I think this code is a little brittle, for example if the API returned a 202 response we would raise an exception because it was not the expected 200? I think this is just being copied from elsewhere so I don't want to blow up this PR over this comment.

ekamin-hpe (changes requested):
You should definitely fix the missing import comment and I'd go ahead and fix the `None` comparisons since that's pretty simple to do. The remainder of my comments are fairly nitpicky.

There are a lot of style things (e.g. spacing between functions, spacing around operators) that you could fix by running a style checker or formatter, but I don't know how important that is going to be, so I didn't mention them all individually.

Overall I'm not a huge fan of the amount of `except Exception` statements in this code but I see they pretty much all go to a generic exception handler function. I still probably would not have done it this way but it's all right I guess.

scripts/operations/configuration/write_root_secrets_to_vault.py
```diff
@@ -38,105 +38,101 @@
...
+config = {
+    'version': 1,
+    'filters': {
+        'exclude_errors': {
+            '()': _ExcludeErrorsFilter
+        }
+    },
+    'formatters': {
+        'file_formatter': {
+            'format': '%(process)d | %(asctime)s | %(name)s | %(pathname)s | %(lineno)s | %(funcName)s | %(levelname)s | %(message)s'
```
mharding-hpe, line 64:
```suggestion
            'format': '%(asctime)s.%(msecs)03d | %(process)d | %(name)s | %(pathname)s | %(lineno)s | %(funcName)s | %(levelname)s | %(message)s',
            'datefmt': '%Y%m%d_%H%M%S'
```

scripts/operations/configuration/python_lib/api_requests.py
mharding-hpe: I agree with you on including the status code in the message. I'll make that change.

I do think we would want to raise an exception if the API gave us an unexpected status code, even if it's just a different 2XX code. Or anyway, we'd want to do that on internal systems, since it would mean an API is not giving us the status code that it advertises, and so we'd want to be aware of that and make sure it isn't going to cause us problems. But I can see the argument that at least on customer systems, if we get a mismatched but still plausibly-successful status code, then we could allow it. In truth, though, that doesn't seem like logic that should be inside this function. I would think the caller would need to decide if, for their particular call, variance would be acceptable.

If this function was going to be more widely used, I guess what I would really want to do is allow the user to specify a set of status codes as acceptable, rather than only a single one. That way the caller could decide how flexible they wanted this to be. For the purposes of this PR, though, I'll just include the status code in the exception. I've opened [CASMINST-5605](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5605) to contemplate the other sort of revisions being discussed here.

scripts/operations/configuration/python_lib/api_requests.py
```diff
@@ -0,0 +1,84 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""Shared Python function library: API requests"""
+
+import logging
```
mharding-hpe, line 26:
```suggestion
import logging
import time
```

scripts/operations/configuration/python_lib/api_requests.py
mharding-hpe: Good catch! I forgot that we don't have a Python linter in this repo, just the Bash linter. Something we should change...

scripts/operations/configuration/write_root_secrets_to_vault.py
mharding-hpe: Completely agree.

scripts/operations/configuration/python_lib/api_requests.py
ekamin-hpe: Sure, that's good with me. I'd probably suggest an unexpected 2XX return code should be a warning or something. I just find it a little odd that we would treat e.g. 400 the same as 202.

ekamin-hpe (approved):
LGTM!

mharding-hpe:
> You should definitely fix the missing import comment and I'd go ahead and fix the `None` comparisons since that's pretty simple to do. The remainder of my comments are fairly nitpicky.
> 
> There are a lot of style things (e.g. spacing between functions, spacing around operators) that you could fix by running a style checker or formatter, but I don't know how important that is going to be, so I didn't mention them all individually.
> 
> Overall I'm not a huge fan of the amount of `except Exception` statements in this code but I see they pretty much all go to a generic exception handler function. I still probably would not have done it this way but it's all right I guess.
> 
> I also didn't want to nitpick too much since a lot of it is just refactoring existing code...

I appreciate all of the feedback. I've fixed the specific issues you noted. I'm going to run it through a linter/formatter as well. 

For my education (as someone who came to Python later in my career), how would you do the 'except Exception' stuff differently? Would you only look for exceptions that you can explicitly anticipate the types of, and let the others bubble up to stop the script? Or would you still try: except: for those, except only do it once at a top level, instead of doing it so many times in each function? That way is definitely much cleaner. The only reason I thought it made more sense to do it at the lower level is that it allowed me to give the user a more specific error message about what was being attempted when the failure happened. Obviously the exception will include a stack trace, and the function names do provide clues about what was going on, but that felt less user friendly. It always seemed to me that end users shouldn't be seeing stack traces.

Since you are certainly more experienced in Python than I am, I'd love to hear your thoughts on this, whether or not this specific PR gets refactored accordingly.

ekamin-hpe:
Regarding catching bare excepts it's just kind of considered bad practice, but to elaborate on it a little:

> Would you only look for exceptions that you can explicitly anticipate the types of, and let the others bubble up to stop the script?

Basically, this. You generally don't want to catch the generic `Exception` type and instead should try to catch more specific exceptions.

In my opinion it's actually okay to print tracebacks in cases when an exception occurs that you really were not expecting to happen, but ideally the code would be robust enough to handle the exceptions that actually *can* occur to keep those cases to a minimum.

Here though I honestly don't think it's THAT big of a deal because all the `except Exception` handling that we are doing is just prettify-ing uncaught exceptions that always result in exiting with an error anyway, it would be a lot worse if we were doing something like this:
```
try:
  do_rest_api_request()
except Exception:
  do_handling_of_an_http_error_and_proceed()
```

> Or would you still try: except: for those, except only do it once at a top level, instead of doing it so many times in each function? That way is definitely much cleaner. 

Yeah, this definitely would be more readable but then it doesn't provide all the context.

See also: https://www.flake8rules.com/rules/E722.html

rustydb:
Not to share false hope, but regarding the exception & linting perhaps this can move to csm-common-library at some point (once we use it, which is soon). Otherwise adding a pylint task to docs-csm wouldn't hurt.

PR #2828, opened 2022-11-17
CASM-3633/CASM-3635: write_root_secrets_to_vault.py: Additional options added

mharding-hpe:
# Description

This PR continues the modifications to the write_root_secrets_to_vault script. This PR contains changes for two tickets.

## [CASM-3633](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3633)

Until now, that script did not provide the user any options to govern how it behaved. Instead it would always read in the root password hash from the /etc/shadow file and it would read in the SSH keys from the /root/.ssh/id_rsa* files.

This PR adds options to:

1. Allow the user to specify the source file for the private SSH key
2. Allow the user to specify the source file for the public SSH key
3. Give the user other options to specify the root password -- provide the plaintext root password via environment variable, provide the hashed password string via environment variable, or be prompted to enter the plaintext root password.

In the case where a root password is provided in plaintext, a shadow file-style SHA512 hash string is generated (just as we do in a number of procedures in the CSM documentation: [example1](https://github.com/Cray-HPE/docs-csm/blob/release/1.3/operations/security_and_authentication/Change_Root_Passwords_for_Compute_Nodes.md) [example2](https://github.com/Cray-HPE/docs-csm/blob/release/1.3/operations/security_and_authentication/Update_NCN_Passwords.md)) from it.

If no arguments are specified, the script behaves as it always has. Thus, because there are no accompanying documentation changes in this PR, to the end user there will not yet be any visible change. Future PRs will address this.

I tested the script on surtur, testing the default path, good path for the other options, as well as bad paths. I also made sure that the hash strings being generated by the script work as intended (as in, when I put them in the shadow file, I am able to authenticate using the password).

None of the code that deals with Vault has changed, so I do not believe anyone with particular Vault knowledge is required to review this part of the PR.

## [CASM-3635](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3635)

The primary change is the addition of arguments to the write_root_secrets_to_vault.py script. Previously the script always set values for the password and both SSH keys. These new arguments add the following options:

- Delete field from root secret in Vault
- Retain current value (if any) from root secret in Vault
- 
Each of those options can be specified independently for either SSH key or the root password hash, in addition to the previous options.

This now allows users to use the tool to make changes to just one of the fields, or to remove the field from Vault (which is a supported option for admins who do not want to enable passwordless SSH or do not want the root password kept in Vault).

The default behavior of the tool remains unchanged.

I also did some refactoring of the tool and its libraries as part of this.

I tested the changes on fanta, both the new options and the previous ones.

The new tool options are not yet being documented. They are being added as part of larger work for [CASM-3367](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3367).

# Checklist Before Merging

### DIFF
--- a/scripts/operations/configuration/write_root_secrets_to_vault.py
+++ b/scripts/operations/configuration/write_root_secrets_to_vault.py
@@ -34,6 +34,7 @@ reads them back to verify that they match what was written.
 """
 
 import argparse
+import copy
 import crypt
 import logging
 import sys
@@ -44,13 +45,17 @@ from python_lib import common
 from python_lib import logger
 from python_lib import vault
 
+from python_lib.types import JSONObject
 
 # Default values
 PRI_KEY_PATH = "/root/.ssh/id_rsa"
 PUB_KEY_PATH = "/root/.ssh/id_rsa.pub"
-
-
 MINIMUM_PW_LENGTH = 8
+
+# Constants
+PASSWORD_HASH_FIELD_NAME = "password"
+PRIVATE_KEY_FIELD_NAME = "ssh_private_key"
+PUBLIC_KEY_FIELD_NAME = "ssh_public_key"
 
 
 def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> None:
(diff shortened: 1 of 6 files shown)


rustydb:
I merged #2820 .

scripts/operations/configuration/write_root_secrets_to_vault.py
```diff
@@ -76,38 +85,129 @@ def get_root_hash_from_etc_shadow() -> str:
...
+        # Simplest case -- we have already read in the hashed value from the environment variable
+        pw_hash = parsed_args.pw_hash_env_var
+    elif parsed_args.pw_env_var is not None:
+        # Generate the hash
+        logging.debug("Generating password hash")
+        pw_hash = crypt.crypt(parsed_args.pw_env_var)
+    elif parsed_args.pw_prompt:
+        # Prompt user for the password
+        logging.debug("Prompting user for password")
+        pw_text = getpass.getpass("Enter desired password: ")
```
ekamin-hpe, line 190: It could be marginally more user friendly to do this in a loop, e.g.:
```
passwords_match = False
while not passwords_match:
  pw_text = getpass.getpass("Enter desired password: ")
  passwords_match = pw_text == getpass.getpass("Verify password: ")
  if not passwords_match:
    logging.error("Passwords do not match. Please try again.")
```

Also definitely not necessary but might be nice to refactor this prompting into a helper function.

ekamin-hpe (changes requested):
Unless I'm missing something, probably should fix that string length checking thing. My other comment is more of an optional suggestion.

ekamin-hpe (approved):
LGTM

scripts/operations/configuration/write_root_secrets_to_vault.py
mharding-hpe: Good call. Done!

ekamin-hpe: Refactored version looks good to me too.

scripts/operations/configuration/write_root_secrets_to_vault.py
```diff
@@ -76,38 +85,138 @@ def get_root_hash_from_etc_shadow() -> str:
...
+def prompt_for_password() -> str:
+    """
+    Prompt the user to enter their desired password string, then
+    have them enter it again. If they do not match, repeat. If they do
+    match, return the password.
+    """
+    logging.debug("Prompting user for password")
+    while True:
+        pw_text = getpass.getpass("Enter desired password: ")
+        if pw_text == getpass.getpass("Verify password: "):
```
ekamin-hpe, line 156: Oh I guess one other question while I'm looking at this, do you want to check anywhere that the password they entered isn't just an empty string? Or any of the other validation stuff like you're doing for the env var case? 

Otherwise it's pretty easy to just hit enter twice to use the empty string as a password. Not sure, maybe that's fine.

mharding-hpe: Done!

mharding-hpe:
(all of my recent force pushes have just been rebases -- I'm done making changes to this PR, absent review comments which necessitate it)

mharding-hpe:
I'm hoping to have a PR up today for [CASM-3635](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3635). Because it depends on the changes in this PR, I'll make the PR against this branch, unless it merges in the meantime. Not a problem either way, just a heads-up that upcoming changes are building on the ones in this PR.

mharding-hpe:
(all of my recent force pushes have just been rebases -- I'm done making changes to this PR, absent review comments which necessitate it)

mharding-hpe:
The shell checker is complaining about scripts which are not part of this PR. This is new behavior.

mharding-hpe:
This is still ready to merge, pending approvals. The shell script checker that is failing is for scripts not related to this PR.

(all of my recent force pushes have just been rebases -- I'm done making changes to this PR, absent review comments which necessitate it)

PR #2831, opened 2022-11-18
CASM-3644 add output of process media stage

dzou-hpe:
# Description

# Checklist Before Merging

### DIFF
--- a/workflows/iuf/operations/extract-release-distributions.yaml
+++ b/workflows/iuf/operations/extract-release-distributions.yaml
@@ -49,8 +49,16 @@ spec:
           productName="{{=jsonpath(inputs.parameters.global_params, '$.product_manifest.current_product.name')}}"
           echo "Extract from: $tarfile for product: $productName"
           dir=$(echo "$tarfile" | sed 's|\(.*\)/.*|\1|')
+          output=${tarfile%.tar.*}
+          echo ${output} > /tmp/parent_directory
+          exit 0 #todo
           cd $dir
           tar -xvzf $tarfile
         volumeMounts:
           - name: iuf
             mountPath: /etc/iuf
+      outputs:
+        parameters:
+          - name: parent_directory
+            valueFrom:
+              path: /tmp/parent_directory
--- a/workflows/iuf/operations/extract-release-distributions.yaml
+++ b/workflows/iuf/operations/extract-release-distributions.yaml
@@ -51,7 +51,6 @@ spec:
           dir=$(echo "$tarfile" | sed 's|\(.*\)/.*|\1|')
           output=${tarfile%.tar.*}
           echo ${output} > /tmp/parent_directory
-          exit 0 #todo
           cd $dir
           tar -xvzf $tarfile
         volumeMounts:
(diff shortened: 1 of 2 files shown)


workflows/iuf/operations/extract-release-distributions.yaml
```diff
@@ -49,8 +49,15 @@ spec:
           productName="{{=jsonpath(inputs.parameters.global_params, '$.product_manifest.current_product.name')}}"
           echo "Extract from: $tarfile for product: $productName"
           dir=$(echo "$tarfile" | sed 's|\(.*\)/.*|\1|')
```
haasken-hpe, line 51: Why not use `dirname`?

```suggestion
          dir=$(dirname "$tarfile")
```

workflows/iuf/operations/extract-release-distributions.yaml
```diff
@@ -49,8 +49,15 @@ spec:
...
+          output=${tarfile%.tar.*}
+          echo ${output} > /tmp/parent_directory
```
haasken-hpe, line 53: Not the biggest deal, but why not call the variable the same as the name of the file it's being stored to?

```suggestion
          parent_directory=${tarfile%.tar.*}
          echo ${parent_directory} > /tmp/parent_directory
```

workflows/iuf/operations/extract-release-distributions.yaml
```diff
@@ -49,8 +49,15 @@ spec:
...
           cd $dir
           tar -xvzf $tarfile
         volumeMounts:
           - name: iuf
             mountPath: /etc/iuf
+      outputs:
+        parameters:
+          - name: parent_directory
```
haasken-hpe, line 61: I also wonder if `parent_directory` is the best name for this or not. Kinda sounds like the parent of the release distribution directory when really it's the extracted release distribution directory. Maybe we can change later if it's confusing.

workflows/iuf/stages.yaml
```diff
@@ -24,106 +24,105 @@
...
-      type: product
-      operations:
-        - name: extract-release-distributions
-          local-path: operations/extract-release-distributions.yaml # this is relative to stages.yaml, which is contained in workflows/iuf/ in docs-csm
-          static-parameters: {} # any parameters that will be supplied statically to this operation.
+  - name: process-media
+    type: product
+    operations:
+      - name: extract-release-distributions
+        local-path: operations/extract-release-distributions.yaml # this is relative to stages.yaml, which is contained in workflows/iuf/ in docs-csm
```
haasken-hpe, line 31: It feels like there should be a schema to this file. It would have multiple benefits:

* Prevent us from repeating documentation on each field throughout this file.
* Ensure it's always valid
* Make it clear to IUF developers how to edit this file and ensure it's still valid based on what the IUF operator expects.

Anyway, nothing for this PR to do since you're just changing whitespace.

haasken-hpe (approved):
Only minor comments. Up to you if you want to address any of them now.

workflows/iuf/operations/extract-release-distributions.yaml
dzou-hpe: agreed, i am just taking whatever is in the scheme def so it won't break operations. we can talk about this monday

leliasen-hpe (approved):

atifsyedali (approved):

PR #2868, opened 2022-11-29
CASM-3681 change extract-release-distrubition to a global stage

dzou-hpe:
# Description

# Checklist Before Merging

### DIFF
--- a/workflows/iuf/operations/extract-release-distributions.yaml
+++ b/workflows/iuf/operations/extract-release-distributions.yaml
@@ -26,6 +26,12 @@ kind: WorkflowTemplate
 metadata:
   name: extract-release-distributions
 spec:
+  tolerations:
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoSchedule"
+  nodeSelector:
+    kubernetes.io/hostname: ncn-m001
   entrypoint: main
   templates:
     - name: main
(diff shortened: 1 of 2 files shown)


rustydb (approved):

workflows/iuf/operations/extract-release-distributions.yaml
```diff
@@ -26,38 +26,80 @@ kind: WorkflowTemplate
...
-          parent_directory=${tarfile%.tar.*}
-          echo ${parent_directory} > /tmp/parent_directory
-          cd $dir
-          tar -xvzf $tarfile
+          set -x
+          media_dir={{inputs.parameters.media_dir}}
+          cd $media_dir
+          tar -xvzf {{inputs.parameters.tarfile}} &> /dev/stderr
+          dir_name=`tar -tzf {{inputs.parameters.tarfile}} | head -1 | cut -f1 -d"/"`
+          cat ${dir_name}/iuf-product-manifest.yaml > /tmp/manifest.yaml
```
haasken-hpe, line 88: This doesn't look like it will fail gracefully if any tar file contained in this directory does not contain an IUF product manifest. I think we want to handle that somehow, don't we?

haasken-hpe:
> Now that we are doing the tar file extraction in the Argo workflow template for this operation, how are we going to validate the IUF Product Manifest files against the schema before passing it along to the subsequent stages?

Ahhh, I see it here: https://github.com/Cray-HPE/cray-nls/pull/110/files#diff-01ac7d5fed370e886b83d6a1cfe917a39ea801bf4a70885d7be24e4dae3327a1R374

workflows/iuf/operations/extract-release-distributions.yaml
dzou-hpe: the operator is dealing with that, plus validation

haasken-hpe: My misunderstanding was based on the fact that I thought this was checking the exit code of the `cat` command. It is not. If the manifest file isn't there, the `/tmp/manifest.yaml` file is still created, but it's empty, and the output parameter is empty. The IUF operator handles that somewhat gracefully here: https://github.com/Cray-HPE/cray-nls/blob/fe275b4dc5a068446a76b360f397683d044a39d0/src/api/services/iuf/sessions.go#L376

This results in kind of a strange error, but it does not break anything. E.g.:

```
2022-11-30T21:28:25.788Z        INFO    iuf/sessions.go:376     manifest: %!s(<nil>) - %!s(<nil>)
```

haasken-hpe (approved):

phalseth-hpe (approved):

PR #2888, opened 2022-12-02
CASM-3635: write_root_secrets_to_vault.py: Add options to clear or retain root password and/or SSH keys in Vault

mharding-hpe:
# Description

**NOTE**: This PR depends on changes in [this yet-to-be-merged PR](https://github.com/Cray-HPE/docs-csm/pull/2828). That is why the PR is against that branch instead of main. I will move this PR to be against main branch once that one merges.

The primary change in this PR is the addition of arguments to the write_root_secrets_to_vault.py script. Previously the script always set values for the password and both SSH keys. These new arguments add the following options:
- Delete field from root secret in Vault
- Retain current value (if any) from root secret in Vault

Each of those options can be specified independently for either SSH key or the root password hash, in addition to the previous options.

This now allows users to use the tool to make changes to just one of the fields, or to remove the field from Vault (which is a supported option for admins who do not want to enable passwordless SSH or do not want the root password kept in Vault).

The default behavior of the tool remains unchanged.

I also did some refactoring of the tool and its libraries as part of this.

I tested the changes on fanta, both the new options and the previous ones.

The new tool options are not yet being documented. They are being added as part of larger work for [CASM-3367](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3367).

# Checklist Before Merging

- [*] If I added any command snippets, the steps they belong to follow the prompt conventions (see [example][1]).
- [*] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].
- [*] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

### DIFF
--- a/scripts/operations/configuration/python_lib/api_requests.py
+++ b/scripts/operations/configuration/python_lib/api_requests.py
@@ -26,7 +26,9 @@
 import logging
 import time
 import traceback
-from typing import Callable
+from typing import Callable, Container, Union
+
+import requests
 
 from . import common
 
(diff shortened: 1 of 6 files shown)


mharding-hpe:
(this is ready for review -- all of the force pushes are just rebases)

PR #2912, opened 2022-12-06
CASMINST-5440 Use license checker from private csm-docker [main]

mtupitsyn:
# Description

The `artifactory.algol60.net/csm-docker` docker registry will become private since January 2023. This change is to provide authentication context for license checking github workflow, which is using container hosted in this registry.

# Testing

Tested by temporarily adding `push:` event to the list of events which trigger workflow:
https://github.com/Cray-HPE/docs-csm/actions/runs/3633719040/jobs/6131051266

### DIFF
--- a/.github/workflows/license-check.yaml
+++ b/.github/workflows/license-check.yaml
@@ -29,15 +29,20 @@ on:
 jobs:
   license-check:
     runs-on: ubuntu-latest
+
+    container:
+      image: artifactory.algol60.net/csm-docker/stable/license-checker:latest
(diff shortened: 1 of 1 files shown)


rustydb (approved):

mtupitsyn:
@ekamin-hpe I think you will need this change to all places where you use `license-check` workflow, for example in `docs-sat`: https://github.com/Cray-HPE/docs-sat/blob/integration/.github/workflows/license-check.yaml

PR #2933, opened 2022-12-12
CASM-3699: Additions to scripts/operations/configuration/python_lib

mharding-hpe:
# Description

Changes to enable upcoming work in [CASM-3714](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3714)

- Refactor some of the modules in `scripts/operations/configuration/python_lib`
- Add modules to obtain VCS credentials and to query the Cray product catalog configmap in Kubernetes

I tested this on frigg.
I regression tested by testing the `write_root_secrets_to_vault.py` script.
I tested the new features at the interactive console and verified that they worked as expected.

# Checklist Before Merging

### DIFF
--- a/scripts/operations/configuration/python_lib/api_requests.py
+++ b/scripts/operations/configuration/python_lib/api_requests.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022-2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -23,6 +23,8 @@
 #
 """Shared Python function library: API requests"""
 
+import base64
+import copy
 import logging
 import time
 import traceback
@@ -31,6 +33,17 @@ from typing import Callable, Container, Union
 import requests
 
 from . import common
+from . import k8s
+from .types import JSONDecodeError
+
+
+ADMIN_CLIENT_SECRET_NAME = "admin-client-auth"
+ADMIN_CLIENT_SECRET_NAMESPACE = "default"
+AUTH_TOKEN_URL = ("https://api-gw-service-nmn.local/"
+                  "keycloak/realms/shasta/protocol/openid-connect/token")
+
+# For type hints
+ApiResponse = requests.models.Response
 
 
 def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> None:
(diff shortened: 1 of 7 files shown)


scripts/operations/configuration/python_lib/api_requests.py
```diff
@@ -48,20 +61,80 @@ def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> N
...
+    admin_client_secret_data = k8s_client.get_secret_data(name=ADMIN_CLIENT_SECRET_NAME,
+                                                          namespace=ADMIN_CLIENT_SECRET_NAMESPACE)
+    try:
+        encoded_client_secret = admin_client_secret_data["client-secret"]
+    except (KeyError, TypeError) as exc:
+        log_error_raise_exception(
+            f"{secret_label} has an unexpected format", exc)
+    # return decoded bytes as a string
+    try:
+        return base64.standard_b64decode(encoded_client_secret).decode()
```
ekamin-hpe, line 78: My IDE complains that `encoded_client_secret` may be referenced before assignment, but that's only because it can't tell that `log_error_raise_exception` always raises an exception. Nothing to necessarily fix here, just wanted to point it out.

scripts/operations/configuration/python_lib/k8s.py
```diff
@@ -32,9 +32,14 @@
 import kubernetes.config
 
 from . import common
+from .types import JSONObject
 
-K8S_CONFIGURATION = None
-K8S_API_CLIENT = None
+# To help for type hinting
+V1Client = kubernetes.client.api.core_v1_api.CoreV1Api
+V1ClientConfiguration = kubernetes.client.configuration.Configuration
```
ekamin-hpe, line 39: I find these first two names a tiny bit confusing, particularly wondering why `V1Client` isn't just `CoreV1Api` instead. As for `V1ClientConfiguration`, I'm less sure but I don't think that one even has a version string. Or maybe just importing them directly instead would be easier.

scripts/operations/configuration/python_lib/k8s.py
```diff
@@ -52,70 +57,107 @@ def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> N
...
-    except Exception as exc:
-        log_error_raise_exception(f"Error retrieving {secret_label}", exc)
+        return k8s_object.data
+    except AttributeError as exc:
+        log_error_raise_exception(
+            f"Error accessing data field in {label}", exc)
 
 
-def get_service(name: str, namespace: str) -> kubernetes.client.models.v1_service.V1Service:
+class Client():
```
ekamin-hpe, line 108: micro nit: parens not needed here

scripts/operations/configuration/python_lib/product_catalog.py
```diff
@@ -0,0 +1,210 @@
+#
```
ekamin-hpe, line 1: There isn't anything inherently wrong with this, but I want to point out that we actually already publish a python client library for the product catalog: https://artifactory.algol60.net/ui/native/csm-python-modules/simple/cray-product-catalog/

Some of the things it does are being duplicated here. Some of the things here are things that the library above does not do. It probably would be more work to correctly set up a dependency on that library.

scripts/operations/configuration/python_lib/vault.py
```diff
@@ -155,134 +137,153 @@ def get_secret_data_from_response(api_response: requests.models.Response) -> JSO
...
+        logging.debug(f"{secret_key} secret URL = {secret_url}")
+        logging.debug(f"Writing {secret_key} secret to Vault")
+        self.api_post(
+            url=secret_url, expected_status_codes=API_STATUS_OK_EMPTY, json=secret_data)
+
+    def delete_csm_root_secret(self, **kwargs) -> None:
+        """
+        Wrapper function that supplies the CSM root secret key to delete_secret()
+        """
+        self.delete_secret(secret_key=CSM_ROOT_SECRET_KEY, **kwargs)
```
ekamin-hpe, line 277:
```suggestion
        self.delete_secret(secret_key=CSM_ROOT_SECRET_KEY)
```

Appears to me `delete_secret` won't accept any extra args.

scripts/operations/configuration/python_lib/product_catalog.py
```diff
@@ -0,0 +1,210 @@
...
+    sorted_version_objects = sorted(version_object_to_string.keys())
+    latest_version_object = sorted_version_objects[-1]
+    # return corresponding string
+    latest_version_string = version_object_to_string[latest_version_object]
+    logging.debug(
+        "get_latest_version: latest_version_string = '%s'", latest_version_string)
+    return latest_version_string
+
+
+class ProductCatalog():
```
ekamin-hpe, line 128: micro nit: parens not needed here

scripts/operations/configuration/python_lib/api_requests.py
ekamin-hpe: There are a lot of instances of this in the code. I'm not worrying about it for now.

scripts/operations/configuration/python_lib/api_requests.py
```diff
@@ -31,6 +33,17 @@
...
+ADMIN_CLIENT_SECRET_NAME = "admin-client-auth"
+ADMIN_CLIENT_SECRET_NAMESPACE = "default"
+AUTH_TOKEN_URL = ("https://api-gw-service-nmn.local/"
+                  "keycloak/realms/shasta/protocol/openid-connect/token")
+
+# For type hints
+ApiResponse = requests.models.Response
 
 
 def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> None:
```
ekamin-hpe, line 49: Is there a reason this function is defined in every module? Can't it just be defined once in `common.py`?

ekamin-hpe (approved):
Looks good to me. Left a few comments, but either nitpick-level or pretty minor.

scripts/operations/configuration/python_lib/api_requests.py
mharding-hpe: Actually, I'm glad you reminded me, because I wanted to ask you about this. The reason I define it locally is because when logging calls are made, they record the source file generating the message. This is meant to be a helper function for logging, and I want it to record the current source file in the log entries, not common.py. Does that make sense? And can you think of an alternative way to accomplish that?

scripts/operations/configuration/python_lib/api_requests.py
mharding-hpe: Yeah, I noticed the same complaint and for the same reason. Do you think it is worth assigning it a null value just to avoid the linting complaint?

scripts/operations/configuration/python_lib/k8s.py
mharding-hpe: Good call!

scripts/operations/configuration/python_lib/product_catalog.py
mharding-hpe: Short term it would be more work, probably, but it's good to know about that. Long term it probably makes sense to not reinvent the wheel here, if possible. Especially since the underlying structure of CPC is supposed to eventually move away from being based on a K8s config map -- I presume if/when that happens, the CPC Python client library will be updated accordingly, so using it would reduce the changes needed here at that time.

scripts/operations/configuration/python_lib/k8s.py
mharding-hpe: Agreed on the naming.

I did it this way (instead of just importing them) because some other modules needed type hints for these things, and I didn't want them to have to do the imports themselves and remember the full paths, requiring them all to be updated if any of them changed. But I can probably do it in a cleaner way. Let me think on it.

scripts/operations/configuration/python_lib/api_requests.py
ekamin-hpe: Nah, I don't think it's worth addressing unless we change how the error handling is done more fundamentally.

scripts/operations/configuration/python_lib/api_requests.py
ekamin-hpe: Maybe it could take the module name as a parameter, e.g. something like this?

```
log_error_raise_exception(msg='We failed', parent_exception=exc, module=__name__)
```

I don't know, that seems kind of awkward. This gets back into the other conversations we've had about error handling as well as the conversation here: https://github.com/Cray-HPE/docs-csm/pull/2933#discussion_r1046438497

But I think normally would I would do in error handling cases is just something like:
```
try:
  something_that_raises_an_error()
except TheError as e:
  LOGGER.error('We got an error ... ')
  raise
```

Then something higher up would handle `TheError`.

I don't feel super strongly about addressing this as you have a good point here, but I suspect that reworking the error handling more fundamentally could help with some of these weird quirks.

scripts/operations/configuration/python_lib/product_catalog.py
mharding-hpe: Captured this in [CASM-3740](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3740)

PR #2964, opened 2022-12-20
MTL-2048 Account for `ARCH` in NCN artifacts

rustydb:
# Description

In lieu of the upcoming aarch64 architecture, all artifacts built by the node-images repository will now denote their architecture. NCNs may never have aarch64 support, but their artifacts are built from the node-images repository.

This change modifies the usage of The NCN artifacts, referring to them with their new names that include their architecture. Without this change, none of the procedures that use the NCN artifacts will be usable.

# Checklist Before Merging

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -534,7 +534,8 @@ if [[ ${state_recorded} == "0" && $(hostname) == "ncn-m001" ]]; then
 
     k8s_done=0
     ceph_done=0
-    if [[ -f ${artdir}/kubernetes/secure-kubernetes-${KUBERNETES_VERSION}.squashfs ]]; then
+    arch="$(uname -i)"
+    if [[ -f ${artdir}/kubernetes/secure-kubernetes-${KUBERNETES_VERSION}-${arch}.squashfs ]]; then
         k8s_done=1
     fi
     if [[ -f ${artdir}/storage-ceph/secure-storage-ceph-${CEPH_VERSION}.squashfs ]]; then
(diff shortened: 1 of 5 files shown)


upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -544,26 +544,26 @@ if [[ ${state_recorded} == "0" && $(hostname) == "ncn-m001" ]]; then
...
 
     set -o pipefail
     IMS_UPLOAD_SCRIPT=$(rpm -ql docs-csm | grep ncn-ims-image-upload.sh)
 
-    export IMS_ROOTFS_FILENAME="${artdir}/kubernetes/secure-kubernetes-${KUBERNETES_VERSION}.squashfs"
-    export IMS_INITRD_FILENAME="${artdir}/kubernetes/initrd.img-${KUBERNETES_VERSION}.xz"
-    export IMS_KERNEL_FILENAME="${artdir}/kubernetes/*.kernel"
+    export IMS_ROOTFS_FILENAME="${artdir}/kubernetes/secure-kubernetes-${KUBERNETES_VERSION}-$(uname -i).squashfs"
+    export IMS_INITRD_FILENAME="${artdir}/kubernetes/initrd.img-${KUBERNETES_VERSION}-$(uname -i).xz"
+    export IMS_KERNEL_FILENAME="${artdir}/kubernetes/*-$(uname -i).kernel"
```
heemstra, line 560: this uname won't give correct output during an upgrade, will it? it'll be the current/old version.

rustydb: Correct me if I'm wrong, but one can not upgrade CSM from one architecture to another architecture.
I think the only way is if the target node was swapped out mid-way through its rebuild.

rustydb: Specifically CSM NCNs.
I don't think we even plan on supporting multi-arch clusters (e.g. arm and x86_64 ceph nodes or k8s nodes mixed together).

jacobsalmela (changes requested):
Just needs those shellchecks fixed.

upgrade/scripts/upgrade/prerequisites.sh
heemstra: yeah, totally. I missed that it was just doing the arch bit and not any version info.

denniswalker (approved):

jacobsalmela (approved):

PR #2986, opened 2023-01-04
CASM-3711: Add function to return NCN xnames from HSM SMD

mharding-hpe:
# Description

This adds a new Python module inside the scripts folder tree. Currently the module is not used, but it will enable future work. At present, the module consists primarily of a single function, which I tested in isolation on mug to verify that it worked as expected.

Some minor linting was also done to other files, to clean up a few minor linter complaints.

# Checklist Before Merging

### DIFF
--- a/scripts/operations/configuration/python_lib/api_requests.py
+++ b/scripts/operations/configuration/python_lib/api_requests.py
@@ -39,8 +39,8 @@ from .types import JSONDecodeError
 
 ADMIN_CLIENT_SECRET_NAME = "admin-client-auth"
 ADMIN_CLIENT_SECRET_NAMESPACE = "default"
-AUTH_TOKEN_URL = ("https://api-gw-service-nmn.local/"
-                  "keycloak/realms/shasta/protocol/openid-connect/token")
+API_GW_BASE_URL = "https://api-gw-service-nmn.local"
+AUTH_TOKEN_URL = f"{API_GW_BASE_URL}/keycloak/realms/shasta/protocol/openid-connect/token"
 
 # For type hints
 ApiResponse = requests.models.Response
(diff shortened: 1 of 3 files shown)


ekamin-hpe:
This is more a comment on the overall parts of `python_lib` that interact with the CSM APIs but I want to point out that we also made a Python package that provides CSM API functions: https://github.com/Cray-HPE/python-csm-api-client

This is pretty similar to the other comment I had about the new ProductCatalog class in #2933, though between the two I bet python-csm-api-client would be easier to adopt than the product catalog API I mentioned then.

There probably would be build issues to figure out though, since the only format python-csm-api-client is published right now is in `pip`-installable binary and source distributions and as I understand it this builds in RPM format and seems to have no dependencies.

mharding-hpe:
> we also made a Python package that provides CSM API functions: https://github.com/Cray-HPE/python-csm-api-client
> 
> This is pretty similar to the other comment I had about the new ProductCatalog class in #2933, though between the two I bet python-csm-api-client would be easier to adopt than the product catalog API I mentioned then.
> 
> There probably would be build issues to figure out though, since the only format python-csm-api-client is published right now is in `pip`-installable binary and source distributions and as I understand it this builds in RPM format and seems to have no dependencies.

I've updated [CASM-3740](https://jira-pro.its.hpecorp.net:8443/browse/CASM-3740) to include this as well.

rustydb (approved):

rustydb:
The more I look at the `https://github.com/Cray-HPE/docs-csm/tree/main/scripts/operations/configuration` folder the more I want us to port this to csm-common-library. That repo should be ready by the end of this week, as such it does have a setup.py file that could pull in @ekamin-hpe's dependency.

ekamin-hpe:
@rustydb interested to hear more about csm-common-library when the time comes. I'm curious if it is also duplicating any of the stuff we implemented in python-csm-api-client.

PR #2994, opened 2023-01-09
CASM-3712/CASM-3713: Create Python functions to get/update BSS bootparameters & CFS configs/sessions

mharding-hpe:
# Description

This PR creates a few Python functions used to query and update BSS bootparameters, and to query and create CFS configurations and sessions. These functions are not currently used by any scripts, but will be used in upcoming PRs.

I tested the BSS functions on mug and validated that they worked as expected. 
I tested the CFS functions on fanta and validated that they worked as expected.
Of course, they will be tested further when they are integrated into other scripts in the future.

# Checklist Before Merging

### DIFF
--- a/scripts/operations/configuration/python_lib/cfs.py
+++ b/scripts/operations/configuration/python_lib/cfs.py
@@ -1,0 +1,150 @@
+#
+# MIT License
+#
+# (C) Copyright 2023 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


scripts/operations/configuration/python_lib/cfs.py
```diff
@@ -0,0 +1,150 @@
...
+                      "expected_status_codes": {200}}
+
+    if not expected_to_exist:
+        request_kwargs["expected_status_codes"].add(404)
+
+    resp = api_requests.get_retry_validate(**request_kwargs)
+    if resp.status_code == 404:
+        # This will only happen if expected_to_exist is set to False and the session
+        # was not found. In this case, return None.
+        return None
```
ekamin-hpe, line 74: I don't think this has any functional effect but the type hint says this function should return a dict.

ekamin-hpe (approved):
LGTM

scripts/operations/configuration/python_lib/cfs.py
mharding-hpe: True. For some reason I got the idea that None was implicitly acceptable as a null value, without being explicitly type hinted. I mean, I know that type hinting isn't enforced, so I could return anything and Python wouldn't bat an eyelash.

I'm not sure where I got the above idea, though, and it's easy enough to make Union types that include None to explicitly inform about this possibility.

mharding-hpe: Fixed

ekamin-hpe: > I know that type hinting isn't enforced

Right, I think it basically doesn't matter. As usual I'm just pointing out things that my IDE feels like pointing out :)

rustydb (approved):
I'm approving albeit a bit blindly, I want all of this to move out of docs once csm-common-library is ready. For now I consider this as "making stuff work" and will refrain from scrutiny.

nrockershousen (approved):

PR #3041, opened 2023-01-20
CASMINST-5675: Create Argo workflow templates for prepare-images

haasken-hpe:
# Description

Create two Argo workflow templates for the two operations in the
`prepare-images` IUF stage, `prepare-management-images` and
`prepare-managed-images`. The only difference between these two stages
is that they reerence a different bootprep input file from the input
parameters on the IUF Activity.

Note that currently these two operations will occur serially, which is
slower than if we did them concurrently. We could probably improve to
get them running concurrently, but it may not always be safe to run two
`sat bootprep` commands concurrently if they are trying to create items
with the same names.

Test Description:
Tested by running with a `sat bootprep` input file that defined a few
simple CFS configurations, an image built from the barebones recipe, a
customized version of that image, and a couple of session templates that
use that image and those configurations.

Should be tested again since a few minor changes were made.

Notably, should be tested again with a `sat bootprep` command that
fails, e.g. on an invalid bootprep file.

# Checklist Before Merging

### DIFF
--- a/workflows/iuf/operations/prepare-images/prepare-managed-images-template.yaml
+++ b/workflows/iuf/operations/prepare-images/prepare-managed-images-template.yaml
@@ -1,0 +1,65 @@
+#
+# MIT License
+#
+# (C) Copyright 2023 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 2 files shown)


jack-stanek-hpe (approved):
Looks good to me.

jesseviola-hpe (approved):

rustydb (approved):

PR #3063, opened 2023-01-23
CASM-3569: Use new default name for NCN personalization CFS configurations; linting

mharding-hpe:
# Description

Most of this is replacing use of the old default NCN personalization CFS configuration name (`ncn-personalization`) with the new one (`management-<csm recipe version>`). In some cases, this necessitated additional changes, usually to reflect the fact that the new default changes with each new recipe.

The apply_csm_config script was changed so that the config-name argument is required, since it can no longer use a simple default value. Pages which called the script were updated to take this into account.

In addition, I removed some related content which was outdated, redundant, and/or superfluous. This PR also includes other minor linting on the same pages.

# Checklist Before Merging


scripts/operations/configuration/apply_csm_configuration.sh
```diff
@@ -177,7 +175,7 @@ while [[ $# -gt 0 ]]; do
 done
 
 ## Set defaults
-[[ -z ${CONFIG_NAME} ]] && CONFIG_NAME=${DEFAULT_CONFIG_NAME}
+[[ -z ${CONFIG_NAME} ]] && err_exit "Configuration name must be specified with --config-name argument"
```
mharding-hpe, line 178:
```suggestion
[[ -z ${CONFIG_NAME} ]] && usage_err_exit "Configuration name must be specified with --config-name argument"
```

operations/configuration_management/NCN_Node_Personalization.md
```diff
@@ -39,7 +39,8 @@ All of the following are prerequisites on the node where this procedure is being
 
 1. (`ncn-m#`) Back up the current CFS state.
```
ekamin-hpe, line 40: I will just point out that I think these instructions will need further changes for 1.4, though probably in a follow-up PR. One of the changes for Alice is to stop using separate NCN personalization and customization configs (i.e., use the same CFS config for both and ensure the specific Ansible plays do the right thing in each context). So going forward we probably won't need steps for editing the file to remove the `ncn-image-customization` config from the input file.

upgrade/Stage_0_Prerequisites.md
```diff
@@ -333,7 +333,7 @@ This upgrade scenario is extremely uncommon in production environments.
     This script will also leave CFS disabled for the NCNs. CFS will automatically be re-enabled on them as they are rebooted during the upgrade.
 
     ```bash
-    /usr/share/doc/csm/scripts/operations/configuration/apply_csm_configuration.sh --no-enable
+    /usr/share/doc/csm/scripts/operations/configuration/apply_csm_configuration.sh --no-enable --config-name management-0.0
```
ekamin-hpe, line 336: Was this supposed to be management-23.03 like the other examples?

ekamin-hpe (approved):
Looks good to me. One minor comment about an "example" value, and something I think will go in a follow-up PR.

operations/configuration_management/NCN_Node_Personalization.md
mharding-hpe: Yup, I figured as much. But like you, I think that's best done in a separate PR.

upgrade/Stage_0_Prerequisites.md
mharding-hpe: No, because this is not using the recipe file. This is the actual command that gets run on CSM internal development team installs. I figured this would be a way to distinguish them. I didn't want to put a value here like 23.03 because I wouldn't want to have to go and update it for each new release.

mharding-hpe: That said, hopefully we can get to a point where we don't need this alternate procedure for internal development installs. Maybe we're already there for 1.4 -- I'm not sure. If that's the case, we'd want to reword this entire section to eliminate the "internal CSM-only install" option.

ekamin-hpe: I see, thanks for sharing the context. I was not realizing this was just in the CSM-only instructions. It might be useful to clarify why that name is being used, even though I'm gathering that the name at this point is somewhat arbitrary.

rustydb (approved):

PR #3065, opened 2023-01-23
CASM-3728: Update product catalog

jsollom-hpe:
# Description

s3_upload updates the product catalog with an array of dictionaries containing the keys: bucket and key where the uploaded artifact has been stored.

It is stored in the product catalog under component_versions like so

  component_versions:
    s3:
    - bucket: boot-images
      key: Analytics/Cray-Analytics.x86_64-base.1.squashfs
    - bucket: boot-images
      key: Analytics/Cray-Analytics.x86_64-base.2.squashfs
    - bucket: boot-images
      key: Analytics/Cray-Analytics.x86_64-base.3.squashfs

# Checklist Before Merging


workflows/iuf/operations/s3-upload/s3-upload-template.yaml
```diff
@@ -72,6 +72,26 @@ spec:
                 PARENT_DIR=$(echo '{{inputs.parameters.global_params}}' | jq -r '.stage_params."process-media".current_product.parent_directory')
 
                 /usr/local/bin/s3-upload "$CONTENT" "$PARENT_DIR"
+    - - name: s3-manufacture-product-catalog-data
+        template: make-product-catalog-content
+        arguments:
+          parameters:
+            - name: global_params
+              value: "{{inputs.parameters.global_params}}"
```
ekamin-hpe, line 80: I think we had talked about this previously but just wanted to bring it up again - it would probably be cleaner to just pass in the s3 content from the current product manifest instead of the entire `global_params` structure, and call it something like `manifest_s3_content`. I think that would look like this:

```suggestion
            - name: manifest_s3_content
              value: "{{=jsonpath(inputs.parameters.global_params, '$.product_manifest.current_product.manifest.content.s3')}}"
```

And the name of the input parameter for `make-product-catalog-content` would have to change from `global_params` to `manifest_s3_content` (line 153).

And finally the script itself would change a little, I think it would be:

```
ARRAY=$(echo '{{inputs.parameters.manifest_s3_content}}' | jq -r '[.[] | {"bucket": .bucket, "key": .key}]')
```

(note: I haven't tested any of this so I could have made a mistake).

ekamin-hpe (commented):
Looks good, though I want to get your take on the change I've suggested.

workflows/iuf/operations/s3-upload/s3-upload-template.yaml
jsollom-hpe: I committed your suggestion, and I'll make other additional modifications re-test it. I'll put the PR into draft state until the tests pass.

workflows/iuf/operations/s3-upload/s3-upload-template.yaml
```diff
@@ -136,3 +156,23 @@ spec:
...
+        valueFrom:
+          path: /tmp/product-content
+    script:
+      image: registry.local/artifactory.algol60.net/csm-docker/stable/docker.io/iuf:v0.1.0
+      command: [bash]
+      source: |
+        ARRAY=$(echo '{{inputs.parameters.global_params}}' | jq -r '[.product_manifest.current_product.manifest.content.s3[] | {"bucket": .bucket, "key": .key}]')
+        if [ $? -ne 0 ]; then
+          ARRAY="[]"
+          echo "No S3 content found."
```
ekamin-hpe, line 176: One thing I happened to notice here, I think it probably makes sense to echo this to `stderr`.

workflows/iuf/operations/s3-upload/s3-upload-template.yaml
ekamin-hpe: I think I figured out how we can make this work, if `value` uses the `toJson` function then I think you get a valid JSON structure that can be passed to the `jq` command in your script.
```
value: "{{=toJson(jsonpath(inputs.parameters.global_params, '$.product_manifest.current_product.manifest.content.s3'))}}"
```

ekamin-hpe (approved):
Looks good. It appears to me that checking the exit code when assigning `ARRAY` will handle the case when the manifest has no s3 content.

IMO it's still a little awkward in the case when a manifest has no s3 content, because what happens is the literal string `"{{=toJson(jsonpath(inputs.parameters.global_params, '$.product_manifest.current_product.manifest.content.s3'))}}"` gets passed into the script that runs `jq`, but I think the error handling covers it.

workflows/iuf/operations/s3-upload/s3-upload-template.yaml
```diff
@@ -136,3 +156,23 @@ spec:
...
+  
+  - name: make-product-catalog-content
+    inputs:
+      parameters:
+      - name: manifest_s3_content
+    outputs:
+      parameters:
+      - name: product-content
+        valueFrom:
+          path: /tmp/product-content
```
ekamin-hpe, line 168: I'm actually going to leave one more suggestion. As an extra layer of protection to ensure that we don't fail when the script unexpectedly fails to write out /tmp/product-content, we can add a default value here.

```suggestion
        valueFrom:
          path: /tmp/product-content
          default: "{}"
```

jsollom-hpe: I fixed this, but not the way Eli suggested.

haasken-hpe (approved):

PR #3080, opened 2023-01-25
CASMNET-2028 | Transceiver and cable guide for management network

kosonenj:
# Description

Transceiver and cable guide for management network

# Checklist Before Merging


gbaker-hpe:
I this the transceiver guide looks really good.  I left a few comments but that information is great and I think it will really help.

I think we should add a second page that better highlights the PoR switch connections and then points the user to the right cables.  For example, We have different PoR scenarios where we have spines and leafs that connect, and various other combinations.  It would be good to include a diagram of an Aruba spine (with model) to an Aruba Leaf (with model) and then list the cables that would work for those runs.

gbaker-hpe (changes requested):
Let's get those hard tabs the linter is calling out fixed.  Also, Rusty is updating the linter CLI versions which should allow tests to be skipped with inline comments which will address the header linting failures.

gbaker-hpe (approved):

kosonenj (commented):
ready to merge

gbaker-hpe:
This has been discussed in Slack by many people and it's good to go.

rustydb (approved):

PR #3122, opened 2023-01-30
CASMINST-5815: add NCN IMS image name and id to Cray Product Catalog

heemstra:
# Description

# Checklist Before Merging

### DIFF
--- a/scripts/operations/node_management/ncn-ims-image-upload.sh
+++ b/scripts/operations/node_management/ncn-ims-image-upload.sh
@@ -64,6 +64,20 @@ if [ -z "$IMS_ROOTFS_FILENAME" ]; then
     exit 1
 fi
 
+if [[ -z ${CSM_RELEASE} ]]; then
+    echo "\$CSM_RELEASE is not specified" >&2
+    exit 1
+fi
+
+if [[ -z ${CSM_ARTI_DIR} ]] && [[ -z ${CSM_PATH} ]]; then
+    echo "One of \$CSM_ARTI_DIR or \$CSM_PATH must be set to the path of unpacked CSM tarball" >&2
+    exit 1
+fi
+
+CSM_TARBALL=${CSM_ARTI_DIR:-$CSM_PATH}
+
+CPC_VERSION=$(find "$CSM_TARBALL"/docker/artifactory.algol60.net/csm-docker/stable/ -maxdepth 1 | awk -F':' /cray-product-catalog-update/'{print $NF}' | sort -V | tail -1)
+
 IMS_ROOTFS_MD5SUM=$(md5sum "$IMS_ROOTFS_FILENAME" | awk '{ print $1 }')
 IMS_INITRD_MD5SUM=$(md5sum "$IMS_INITRD_FILENAME" | awk '{ print $1 }')
 IMS_KERNEL_MD5SUM=$(md5sum "$IMS_KERNEL_FILENAME" | awk '{ print $1 }')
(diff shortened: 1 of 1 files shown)


scripts/operations/node_management/ncn-ims-image-upload.sh
```diff
@@ -122,4 +136,16 @@ cray ims images update "$IMS_IMAGE_ID" \
...
         --link-path "s3://boot-images/$IMS_IMAGE_ID/manifest.json" > /dev/null
 
+podman run --rm --name ncn-cpc \
+    --user root \
+    -e PRODUCT=csm \
+    -e PRODUCT_VERSION=$CSM_RELEASE \
+    -e YAML_CONTENT_STRING="{images: {$IMS_IMAGE_NAME: {id: $IMS_IMAGE_ID}}}" \
+    -e KUBECONFIG=/.kube/admin.conf \
+    -e VALIDATE_SCHEMA="true" \
+    -e SET_ACTIVE_VERSION="true" \
```
ekamin-hpe, line 146: I'll suggest that we not use the `SET_ACTIVE_VERSION` flag here. I don't think we would want to be changing what version of csm is "active" at this point.

We've decided to move away from using this flag entirely (lots of context found in: https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1629) but in any case I don't think we would want the value changing just as a part of uploading the ims images.

scripts/operations/node_management/ncn-ims-image-upload.sh
```diff
@@ -122,4 +136,16 @@ cray ims images update "$IMS_IMAGE_ID" \
...
+    -v /etc/kubernetes:/.kube:ro \
+    -v ${PWD}/catalog:/catalog:ro \
```
ekamin-hpe, line 148: If you are passing the YAML content as a string rather than a file, I don't think there is any need to mount the working directory into the container.

ekamin-hpe (approved):
Looks good. I think a couple of the env vars are unnecessary though.

scripts/operations/node_management/ncn-ims-image-upload.sh
heemstra: Thanks - left over from testing.  Removing...

scripts/operations/node_management/ncn-ims-image-upload.sh
```diff
@@ -122,4 +136,14 @@ cray ims images update "$IMS_IMAGE_ID" \
         --link-etag "${MANIFEST_ETAG}" \
         --link-path "s3://boot-images/$IMS_IMAGE_ID/manifest.json" > /dev/null
 
+podman run --rm --name ncn-cpc \
+    --user root \
+    -e PRODUCT=csm \
+    -e PRODUCT_VERSION=$CSM_RELEASE \
+    -e YAML_CONTENT_STRING="{images: {$IMS_IMAGE_NAME: {id: $IMS_IMAGE_ID}}}" \
```
haasken-hpe, line 143: I'm not sure how this would behave, but you might want to quote `$IMS_IMAGE_NAME` in this string in case it contains any weird characters. That is, maybe escape double quotes inside this string. I think that would ensure that the key is always valid and of string type. You might want to do the same for `IMS_IMAGE_ID` too.

haasken-hpe (approved):

scripts/operations/node_management/ncn-ims-image-upload.sh
heemstra: Makes sense.  I tried it on surtur to make sure there were no unexpected issues.

PR #3265, opened 2023-02-08
CASMINST-5942: Modify NCN image customization procedure to be more flexible

mharding-hpe:
# Description

The 1.3 procedure for customizing NCN images is written assuming that the S3 paths for the images will always be of the form `k8s/<version>/<artifact>` or `ceph/<version>/<artifact>`. However, that will only be the case the first time the images are customized. As part of the customization process, the new images being used will have their S3 paths of the form `<IMS ID>/<artifact>`. Because of this, the procedures as written do not work for those re-customizations of images that had been previously customized.

This PR modifies the procedures to be more general, not making an assumption about which form the S3 path must take. Almost all of this involves just changing the initial step to identify and download the NCN image artifacts. Minor changes are needed to other parts of the procedure to reflect some of the changes made in that earlier step.

I manually tested the updates steps in isolation, but I have not had the opportunity to test the full procedure on an installed system. However, given the limited scope of the changes, I believe the risk is tolerable.

In addition to the above, I also did some minor linting work, including adding some links back to the glossary for some of the acronyms that previously had not been defined on the page.

# Checklist

### DIFF
--- a/operations/configuration_management/Management_Node_Image_Customization.md
+++ b/operations/configuration_management/Management_Node_Image_Customization.md
@@ -1,7 +1,8 @@
 # Management Node Image Customization
 
-**NOTE:** Some of the documentation linked from this page mentions use of the Boot Orchestration Service (BOS). The use of BOS
-is only relevant for booting compute nodes and can be ignored when working with NCN images.
+**NOTE:** Some of the documentation linked from this page mentions use of the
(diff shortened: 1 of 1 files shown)


rustydb (approved):

nrockershousen (approved):

mharding-hpe:
@don-bahls-hpe please review when you have the opportunity, to make sure that this addresses your concerns.

don-bahls-hpe (approved):

PR #3289, opened 2023-02-13
CASMTRIAGE-4824 make install_csi hook more robust

leliasen-hpe:
# Description

Make install_csi hook more robust. It will check both sp3 and sp4 for link to install csi and will only fail if unable to install csi, independent of installing other repositories. Provides an error message if nexus appears to be down.

# Checklist


rustydb:
Feedback on the backport also applies here: https://github.com/Cray-HPE/docs-csm/pull/3290#pullrequestreview-1296379186

gbaker-hpe (changes requested):
Let's reduce the logic checking in this to avoid having to maintain it when we move so SP5 and beyond.  We can ensure the CSI build publishes to all of these repos and in the future, create a non-specific repo for packages that can be used in all distros.

denniswalker (approved):

PR #3302, opened 2023-02-15
CASMINST-5988

atifsyedali:
# Description

[CASMINST-5988](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5988) BREAK/FIX: management-node-rollout breaks with "default" product containing jinja templates

# Checklist


workflows/iuf/stages.yaml
```diff
@@ -83,14 +83,17 @@ stages:
         static-parameters: {} # any parameters that will be supplied statically to this operation.
         required-manifest-attributes:
           - content.vcs
+        include-default-product-in-site-params: true # this operation wants to receive site_params.products.default in supplied global_params
```
haasken-hpe, line 86: So for all the other operations, this defaults to `false` somehow, right?

haasken-hpe (approved):

workflows/iuf/stages.yaml
haasken-hpe: I confirmed with Atif earlier today that this does default to false.

> Ryan: how does defaulting the include-default-product-in-site-params work? Is it just using the default value for an empty value of type bool in golang?
> Atif: yes. default is false

rustydb (approved):

PR #3299, opened 2023-02-14
CASMINST-5979: Correct procedures for customizing NCN image timezone post-install

mharding-hpe:
# Description

The procedure to update NCN timezones has a few errors in it. In addition, after the CSM install is done, it really should only be used just for timezones, and passwords/SSH keys should be handled using CFS. This PR overhauls that procedure (and makes some small script tweaks) to accomplish the above. While writing and testing these changes, I found other odds and ends in related procedures that I also cleaned up with this PR.

I tested this on fanta except for actually updating the entries in BSS (but I did test the commands to do that -- I just had them echo rather than run the final command to do the update).

- [1.4 backport PR](https://github.com/Cray-HPE/docs-csm/pull/3310)
- [1.3 backport PR](https://github.com/Cray-HPE/docs-csm/pull/3311)

The PRs should be reviewed by at least one of the following folks before merging: @heemstra , @haasken-hpe , @deantroe , @don-bahls-hpe 

Notes/caveats:

- I eventually want to replace the manual code being used to update the BSS entries in this procedure. I wrote up a script to do it (just a more careful version of the basic shell commands being done), but I think that may be better off in a separate PR (just because this one is already getting on the large side). I know that the **csi** tool has that ability, but unless we're going to start making that available on the NCNs by default, I'd like to move away from relying on it for operational procedures (see [CASMINST-4784](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-4784)).
- In a recent discussion with @haasken-hpe and @leliasen-hpe , we identified further changes which are going to be needed to procedures in this PR, as well as related procedures. However, more thought and planning is required before we make those. 

Given the above two points, some feedback for this PR may get shunted into a future ticket, since we know more work is required. The most important thing that this PR needs to be accomplishing is to fix the broken timezone procedure.

# Checklist

- [X] If I added any command snippets, the steps they belong to follow the prompt conventions (see [example][1]).
- [X] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].
- [X] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

 
[1]: https://github.com/Cray-HPE/docs-csm/blob/main/introduction/documentation_conventions.md#using-prompts
[2]: https://github.com/Cray-HPE/teams

PR #3311, opened 2023-02-16
CASMINST-5979: Correct some NCN customization procedures

mharding-hpe:
Backport of https://github.com/Cray-HPE/docs-csm/pull/3299

This includes backporting just the IMS upload portion of the ncn-ims-image-upload.sh script (and not the cray-product-catalog part of it), so that it can be used to simplify some of the procedures.

### DIFF
--- a/operations/configuration_management/Management_Node_Image_Customization.md
+++ b/operations/configuration_management/Management_Node_Image_Customization.md
@@ -36,66 +36,81 @@ The Cray command line interface must be configured on the node where the command
       ARTIFACT_S3_PREFIX="k8s/${ARTIFACT_VERSION}"
       ```
 
-    * If not doing a CSM upgrade, and if the image to be modified is the image currently booted on an NCN, then identify the image
-      by examining the boot parameters used to boot the NCN in question.
-
-      1. Set the `BOOTED_NCN` variable to the hostname of the NCN that is booted using the image that is to be modified.
-
-         > For example, `ncn-w001` or `ncn-s002`.
-
-         ```bash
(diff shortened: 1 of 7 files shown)


operations/configuration_management/Management_Node_Image_Customization.md
```diff
@@ -36,66 +36,81 @@ The Cray command line interface must be configured on the node where the command
...
+      1. Extract the S3 path prefix for the image used to boot the chosen NCN.
 
-         ```bash
-         ARTIFACT_S3_PREFIX=$(cray bss bootparameters list --name "${NCN_XNAME}" --format json | \
-                              jq -r '.[0].params' | \
-                              sed 's#\(^.*[[:space:]]\|^\)metal[.]server=[^[:space:]]*/boot-images/\([^[:space:]]\+\)/rootfs.*#\2#')
-         echo "${ARTIFACT_S3_PREFIX}"
-         ```
+          ```bash
+          ARTIFACT_S3_PREFIX=$(ssh "${BOOTED_NCN}" sed \
```
heemstra, line 99: This won't yield the expected result if $BOOTED_NCN is currently booted from disk. When booted from disk, there is no metal.server parameter.

heemstra (changes requested):

operations/configuration_management/Management_Node_Image_Customization.md
mharding-hpe: Good call. I'll add that caveat as well (in addition to the PIT boot note)

haasken-hpe (approved):

heemstra (approved):

PR #3344, opened 2023-02-24
CASMINST-6009 Slight change in expected output of logwrite_root_secrets_to_vault.py

SavioJoshua:
# Description

# Checklist


rustydb (approved):

gbaker-hpe (changes requested):
@ganeshs-hpe Do you know why we're querying Vault using HTTP (not HTTPS) and an IP instead of a hostname?

The IP opens us up to man in the middle attacks and HTTP is not a secure transport which is important since an SSH private key is being transmitted.

I know you copied what was already written, but do you know why this is setup this way?  Is it just the output that shows HTTP?

I think we want to evaluate this script before we proceed to ensure we're not doing something that is worth correcting.

denniswalker (approved):

PR #3361, opened 2023-02-27
CASMTRIAGE-4865 - Ensure switches have SNMP configured (Cheshire Cat)

gbaker-hpe:
jira-ticket: "[CASMTRIAGE-4865](https://jira-pro.its.hpecorp.net:8443/browse/CASMTRIAGE-4865)"
csm-version: "1.5"
shasta-version: "Cheshire Cat"
---

The CSM install and upgrade now explicitly ask admins to ensure that SNMP is configured on the management network switches so that REDS Hardware Discovery will be successful.

Added information about CANU custom configs to the CSM documentation as a reference for SNMP configurations.

Generally improved SNMP documentation information and backlinking.


gbaker-hpe:
We have two other pages that need Mellanox SNMP examples.

* https://github.com/Cray-HPE/docs-csm/blob/main/operations/network/management_network/backup_custom_configurations.md
* https://github.com/Cray-HPE/docs-csm/blob/main/operations/network/management_network/mellanox/snmpv3_users.md

lukebates123 (approved):

trad511 (changes requested):
We should give custom config examples of snmp configs for Dell and Aruba for leaf-bmc:

```snmp-server vrf default
snmpv3 user testuser auth md5 auth-pass ciphertext deadbeef== priv des priv-pass ciphertext moredeadbeef==``` for aruba example

**Review by trad511 — APPROVED** [review]

**Review by trad511 — APPROVED** [review]

File: install/prepare_site_init.md [hunk 1]
```diff
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable MD013 -->
```
rustydb, line 1: These are very unfriendly to terminals.

rustydb (changes requested):
There's a bunch of sneaky overrides for line-length in here.

Also a bunch of .spelling file changes that seem to disrupt the A-Z sorting, but I see lots of other problems there unrelated to this PR so I'll let them be.

I'll use an auto-formatter and replace the overrides for the line length.

lukebates123 (approved):

install/prepare_site_init.md
gbaker-hpe: Since we include image assets throughout the docs, I would argue that the docs are not designed for consumption within a terminal.

gbaker-hpe: What am I missing? Vim and less both wrap text and the bulk of our text isn't code, it's copy in paragraphs. We also include image assets in the docs without alt text. If we wanted to target the terminal, wouldn't we be post-processing the markdown and rendering it into txt documents?

Don't we want to eliminate extra newlines so that we can format and stylize the text properly in the PDF and HTML versions of the docs without the syntax driving the formatting?

rustydb (changes requested):

install/prepare_site_init.md
rustydb: We moved this conversation to offline, and the consensus was to let the docs-team discuss disabling this check globally for docs-csm. We don't see the newlines affecting the hugo rendered docs, but in general we aren't aware of non-wrapped usage of the raw doc files (where the line length is helpful).

rustydb (approved):

PR #3404, opened 2023-03-06
CASMTRIAGE-5030 Remove local ceph images before node upgrade 

leliasen-hpe:
# Description

Remove local ceph images before node upgrade. Do not run pre-load-images, instead pull images from nexus. This makes sure there are not multiple sha values for the same image stored locally. There should only be one sha value used which is the sha in nexus.

# Checklist


workflows/ncn/storage/storage.upgrade.yaml
```diff
@@ -240,6 +252,62 @@ spec:
...
+                      ceph orch host maintenance enter $TARGET_NCN --force
+                      if [[ $? -ne 0 ]]; then
+                        echo "ERROR setting maintenance mode on $TARGET_NCN"
+                        exit 1
+                      fi
+                    fi
+                    # remove local images on target_ncn
+                    echo "Removing local images on $TARGET_NCN"
+                    ssh $TARGET_NCN podman rmi --all -f
+                    #exit maintenance mode on target_ncn
```
cdelatte-hpe, line 276: we want to pull the ceph image prior to exiting maintenance mode.  the monitoring stack can be addressed like below, but we should do a pull and confirm the sha since it will have changed from when we pushed it.

cdelatte-hpe (changes requested):
might be not necessary, but I have sneaking suspicion this will work and eventually we may hit a scenario where we set have the wrong sha and ceph won't start.

cdelatte-hpe (approved):

djlapenta (approved):

dzou-hpe (approved):

PR #3474, opened 2023-03-17
CASMINST-6090 get critical pod by its container name instead of a grep

dzou-hpe:
CASMINST-6090 get critical pod by its container name instead of a grep
CASMINST-6092 fix broken IUF templates

# Description

# Checklist


workflows/ncn/hooks/before-all/move-critical-singleton-pods.yaml
```diff
@@ -52,9 +52,10 @@ spec:
 
     echo
 
-    pods=( "nexus" "ipxe" "kea" "cray-cfs-api-db" )
-    for pod in ${pods[@]};do
-      podName=$(kubectl get pods --all-namespaces | awk '{print $2}' | grep $pod)
+    containerNames=( "nexus" "ipxe" "kea" "cray-cfs-api-db" )
+    for containerName in ${containerNames[@]};do
+      # get pod name by container name
+      podName=$(kubectl get pods -A -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.status.phase}{"\t"}{range .spec.containers[*]}{.name}{"\t"}{end}{"\n"}{end}' | grep -E "^[^[:space:]]+[[:space:]]+Running[[:space:]]+.*${containerName}.*$" | cut -f1)
```
johren-hpe, line 58: This is too complicated and hard to follow.  It is common kubectl practice to use labels for this specific situation rather than grep/awk/cut.

For example:
```
ncn-m001:~ # kubectl get pods -A -l app.kubernetes.io/name=cray-cfs-api-db -o custom-columns=PodName:.metadata.name --no-headers=true
cray-cfs-api-db-5fddcf7864-m6qsk
```

johren-hpe (changes requested):

workflows/ncn/hooks/before-all/move-critical-singleton-pods.yaml
johren-hpe: The approach above will still catch cases where a pod is unexpected created that has a container with "nexus" or "ipxe" or "kea" as a substring of the container name.

johren-hpe: https://kubernetes.io/docs/reference/labels-annotations-taints/

dzou-hpe: while i agree we should use label but i'd disagree that we should use any of our existing labels. We should just have `critical-pod:true` so i can also remove line 55. any pod requires a move during upgrade should declare that.

johren-hpe: I'm not following why we shouldn't use our existing labels.   For this set of pods we have labels that are specific only to those pods.   I'm okay if you want to add a {{critical-pod:true}} label at some point to be able to get all of the critical pods with one label.   But that is going to take some time to get those labels added and we won't be able to rely on that for older pods.   Using the labels that are there now will be much better than doing the greps that are being done right now.

workflows/ncn/hooks/before-all/move-critical-singleton-pods.yaml
```diff
@@ -52,24 +52,20 @@ spec:
...
-      
-      currentNode=$(kubectl get po -A -o wide | grep $podName | awk '{print $8}')
-      if [[ "$nodeMoveTo" != "$currentNode" ]]; then
-        echo "Move Pod: $podName to Node: $nodeMoveTo"
-        /opt/cray/platform-utils/move_pod.sh $podName $nodeMoveTo > /dev/null
-      fi
+    criticalPodLables=( "app=nexus" "app.kubernetes.io/name=cray-ipxe" "app.kubernetes.io/name=cray-dhcp-kea" "app.kubernetes.io/name=cray-cfs-api-db" )
+    for label in ${criticalPodLables[@]};do
+      pods=($(kubectl get pods -A -l ${label} -o custom-columns=:.metadata.name))
+      # we may have multiple pods matched with the lable
```
johren-hpe, line 58: The name of this script is "move-critical-singleton-pods".  Should we log or fail if we find multiple pods in this case since that is unexpected?

dzou-hpe: i think i will rename the file? it will be future proofing when we want to add specific label  like `critical:true`.

rustydb (approved):

nrockershousen (approved):

PR #3512, opened 2023-03-24
CASMINST-6132: Create/update scripts to handle IMS ID and S3 etag replacement in BOS session templates

mharding-hpe:
# Description

The IMS import procedure ends up changing the IMS IDs of the recipes and images that it imports. In addition, some S3 etag values are changed.
The automated procedure for the IMS import generates a JSON file with a mapping from the old IMS IDs and S3 etags to the new ones.
This PR creates and updates scripts to enable updating BOS with these new IDs. Specifically:
1. The `import_bos_sessiontemplates` tool is updated to allow the user to pass in the JSON ID/etag map file generated by the IMS import. If specified, it will automatically convert any IMS IDs and S3 etags in the session templates as they are imported.
2. A new tool (`update_ims_ids_in_bos`) is created to allow the user to update the current session templates in BOS to apply the changes from the JSON ID map file. This would be useful in cases where BOS was restored without using the option described in the last bullet, or in cases where IMS was wiped out and restored but BOS data was not lost.

Several helper modules were created or updated to support the above two items.

This PR does NOT contain any documentation updates to use these new/update tools. Those will come in a separate PR. This is not a problem because the tools which were updated are still completely backwards compatible and will work fine as they are currently documented.

I tested this on wasp and verified that:
1) BOS export still works as expected
2) BOS import still works as expected with no IMS ID map file specified
3) BOS import works as expected when an IMS ID map file is specified

- [1.4 backport](https://github.com/Cray-HPE/docs-csm/pull/3535)
- [1.3 backport](https://github.com/Cray-HPE/docs-csm/pull/3536)

No earlier backports needed as this procedure did not exist prior to 1.3.

# Checklist

- [X] https://github.com/Cray-HPE/docs-csm/pull/3528 has merged

 

rustydb:
Speaking on behalf of the libCSM development team and docs-CSM review team, the new Python files added here are o.k. because they're apart of a pre-existing set of Python scripts. Migrating these are going to be apart of a larger effort.

mharding-hpe:
> Speaking on behalf of the libCSM development team and docs-CSM review team, the new Python files added here are o.k. because they're apart of a pre-existing set of Python scripts. Migrating these are going to be apart of a larger effort.

Yeah, I'd rather this happen in libcsm now, but there are time constraints, so migrating them later must suffice

mharding-hpe:
Converting to draft because I just realized that BOS session templates also reference S3 etag values, and I believe those will also change and should be updated.

scripts/operations/system_recovery/bos_cli.py
```diff
@@ -0,0 +1,110 @@
...
+        # Make sure the data has been written to the file so the CLI command can read it.
+        tmp.flush()
+        if bos_version == 1:
+            create_command = ["cray", "bos", f"v{bos_version}", "sessiontemplate", "create",
+                              "--file", tmp.name, "--name", template_name, "--format", "json"]
+        else:
+            create_command = ["cray", "bos", f"v{bos_version}", "sessiontemplates", "create",
+                              "--file", tmp.name, template_name, "--format", "json"]
+        try:
+            subprocess.run(create_command, stdout=subprocess.PIPE, check=True)
```
rbak-hpe, line 64: Is there a reason to be using the cli rather than api calls here?  Typically we recommend against that for scripts.

mharding-hpe: I only did this because it's what the existing BOS and IMS import/export scripts did. I agree with you that using the API makes more sense. For now, since I was already making a lot of other changes, I didn't want to make this change on top of that. But at some point I meant to follow up and find out who on the team authored the existing scripts and see why they chose to use the CLI instead of the API. Since it was done for both IMS and BOS, it seemed unlikely to be a coincidence, given that different people on the team owned those.

jsollom-hpe:
> Converting to draft because I just realized that BOS session templates also reference S3 etag values, and I believe those will also change and should be updated.

I don't think the etag values are super important. If they're wrong, BOS issues a warning, and it's not a big deal. I think we might even be able to let them be blank, but that would require further testing.

mharding-hpe:
> > Converting to draft because I just realized that BOS session templates also reference S3 etag values, and I believe those will also change and should be updated.
> 
> I don't think the etag values are super important. If they're wrong, BOS issues a warning, and it's not a big deal. I think we might even be able to let them be blank, but that would require further testing.

Well, I already wrote the code to update them. :) And unless you object, it's probably better to update them than to leave them wrong.

mharding-hpe:
This can merge pending approval by @jsollom-hpe . He says he plans to review today.

PR #3537, opened 2023-03-28
ALT-194 Add references to WLM recovery procedures

dgloe-hpe:
# Description

Add references to the WLM restore procedures from the CPE installation guide.

# Checklist

### DIFF
--- a/operations/System_Recovery.md
+++ b/operations/System_Recovery.md
@@ -10,6 +10,8 @@ The following workflows are available:
   * [Keycloak](#keycloak)
   * [Spire](#spire)
   * [Nexus](#nexus)
+  * [Slurm](#slurm)
+  * [PBS](#pbs)
 * [System recovery after fresh install](#system-recovery-after-fresh-install)
 
 ## Prerequisites
@@ -573,5 +575,18 @@ The latest CSM documentation has been installed on the master nodes. See [Check 
 1. (`ncn-mw#`) Restore the critical data.
 
    See [Nexus Export and Restore](package_repository_management/Nexus_Export_and_Restore.md)
+
+### Slurm
+
+To restore Slurm data from backup, follow sections
+*10.3.11 Restore Slurm Accounting Database from Backup* and
+*10.3.13 Restore Slurm Spool Directory from Backup* in the
+**HPE Cray Programming Environment Installation Guide: CSM on HPE Cray EX Systems (S-8003)**.
+
+### PBS
+
+To restore PBS Professional data from backup, follow section
+*10.7.7 Restore PBS home directory from backup* in the
+**HPE Cray Programming Environment Installation Guide: CSM on HPE Cray EX Systems (S-8003)**.
 
 ## System recovery after fresh install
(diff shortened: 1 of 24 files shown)


operations/System_Recovery.md
```diff
@@ -574,4 +576,57 @@ The latest CSM documentation has been installed on the master nodes. See [Check
...
+*10.3.11 Restore Slurm Accounting Database from Backup* and
+*10.3.13 Restore Slurm Spool Directory from Backup* in the
+**HPE Cray Programming Environment Installation Guide: CSM on HPE Cray EX Systems (S-8003)**.
+
+After restoring Slurm data from backup, check that the procedure was successful.
+
+1. (`uan#`) Check the accounting records.
+
+   ```bash
+   sacct -S <date>
```
kimjensen-hpe, line 591: Would `sacct -a` be a simpler check?  How would one determine the <date> to use if checking foraccount starttimes? 
Also is seems that the <date> is based on CDT+12??  Maybe for those the use slurm this is all a known - so disregard if you feel customers will know how to select a valid <date> for this check.
```
uan01:~ # date
Fri 31 Mar 2023 05:52:54 PM CDT
uan01:~ # sacct -S 2023-03-31T17:55:00
sacct: error: Start time (2023-03-31T17:55:00) requested is after end time (2023-03-31T17:53:15).
uan01:~ # date
Fri 31 Mar 2023 05:53:17 PM CDT
uan01:~ # sacct -S 2023-03-31T17:53:00
JobID           JobName  Partition    Account  AllocCPUS      State ExitCode
------------ ---------- ---------- ---------- ---------- ---------- --------
692208         msgget01      workq                    32    RUNNING      0:0
```

operations/System_Recovery.md
```diff
@@ -574,4 +576,57 @@ The latest CSM documentation has been installed on the master nodes. See [Check
...
+   ```
+
+1. (`uan#`) Check queued jobs.
+
+    ```bash
+   squeue
```
kimjensen-hpe, line 597: If the backup was not successful, would `squeue` be empty?  Is there an example that can be added to show what might be seen if the restore was not successful?  Same question for `sinfo `below` as well as the pbs commands.

dgloe-hpe: The queue should match the state when it was backed up. So if the queue was empty, it should still be empty. The only way you could tell that a restore failed would be if there were jobs in the queue when it was backed up, but they don't show up after the backup/restore.

operations/System_Recovery.md
dgloe-hpe: `sacct -a` will only show the jobs run during the current day. If the backup was done more than a day before the restore, it won't show any jobs. To make sure the backup/restore worked, you want to show jobs from before the backup was made.

kimjensen-hpe: would it make sense to then say that the <date> should be from before the backup was taken?  And what TZ is that date based on?

kimjensen-hpe (approved):

PR #3564, opened 2023-04-02
CASMINST-6150: Update Cray Product Catalog after IMS import

mharding-hpe:
# Description
Starting in CSM 1.4, the Cray Product Catalog contained pointers to IMS image and recipe IDs. This PR modifies the IMS import script so that after import, if any IDs have changed, the Product Catalog will be updated if necessary.

I also provide a standalone script that will just update a specified IMS ID in the Cray Product Catalog. It uses the same backend code, but is intended for use in the docs when a user is manually importing just a single image or recipe, and wants to update the product catalog with the changed ID.

This involved writing code to find the latest version of the product catalog updater image. Code had already been written for parsing the product catalog. This makes use of that, searching each entry in the product catalog for images and recipes. And then if there are any changes for a given CPC entry, the aforementioned CPC updater is used to make the changes (same as is done to set them initially in the NCN image upload script).

I tested the code to update the cray product catalog on wasp, and verified that all of the desired updates were made, and no additional undesired changes were made.

In addition,[this PR relies in [code changes in this PR](https://github.com/Cray-HPE/docs-csm/pull/3585). ***This PR should not merge until that PR merges.***

# Checklist

- - [ ] https://github.com/Cray-HPE/docs-csm/pull/3512 has merged

 
### DIFF
--- a/scripts/operations/configuration/python_lib/product_catalog.py
+++ b/scripts/operations/configuration/python_lib/product_catalog.py
@@ -23,13 +23,15 @@
 #
 """Shared Python function library: Cray product catalog"""
 
+import subprocess
 import traceback
-from typing import Dict, ItemsView
+from typing import Dict, ItemsView, List, Tuple
 
 import logging
 import packaging.version
 import yaml
 
+from . import api_requests
 from . import common
 from . import k8s
 
@@ -37,11 +39,12 @@ from .types import JSONObject
 
 K8S_NAMESPACE = "services"
 K8S_CONFIG_MAP_NAME = "cray-product-catalog"
+CPC_UPDATE_IMAGE = "artifactory.algol60.net/csm-docker/stable/cray-product-catalog-update"
 
 # Used for type hinting
 ProductVersionMap = Dict[str, JSONObject]
 ProductCatalogMap = Dict[str, ProductVersionMap]
-
+NameIdTupleList = List[Tuple[str, str]]
 
 def log_error_raise_exception(msg: str, parent_exception: Exception = None) -> None:
     """
@@ -78,12 +81,11 @@ def parse_product_yaml(label: str, yaml_string: str) -> JSONObject:
             f"YAML data from {label} has an unexpected format", exc)
 
 
-def get_latest_version(product_version_map: ProductVersionMap,
-                       strict_versioning: bool = False,
-                       ignore_invalid_versions: bool = True) -> str:
+def get_latest_version_from_list(version_string_list: List[str],
+                                 strict_versioning: bool = False,
+                                 ignore_invalid_versions: bool = True) -> str:
     """
-    Given a product version map from the product catalog, sort the version strings
-    and return the latest one.
+    Given a list of version strings, sort them and return the latest one.
 
     If strict_versioning is false, then the generic packaging.version.parse() function will be
     used to parse the version strings. Otherwise the packaging.versions.Version() constructor will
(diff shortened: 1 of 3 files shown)


rbak-hpe (approved):

mharding-hpe:
> > I am still testing this, but I wanted to let people start to review it. THIS IS NOT READY TO MERGE
> 
> Is this still true, @mharding-hpe? If so, you might want to put a "WIP: " prefix on it or convert to draft mode to ensure it doesn't get merged prematurely.

I didn't use draft mode because it has been my experience that it's harder to get folks to review those. But I've marked it WIP. I am gratified that Rusty is usually pretty good about not merging hastily, though. :)

scripts/operations/configuration/python_lib/product_catalog.py
```diff
@@ -208,3 +227,101 @@ def get_latest_csm_cfs_information(self) -> Dict[str, str]:
...
+            image_name_ids_content_string_list = [ '"%s": {id: "%s"}' % (image_name, image_id)
+                                                   for image_name, image_id in image_name_ids ]
+            image_name_ids_content_string = ", ".join(image_name_ids_content_string_list)
+            content_string_list.append("images: {%s}" % image_name_ids_content_string)
+        if recipe_name_ids:
+            recipe_name_ids_content_string_list = [ '"%s": {id: "%s"}' % (recipe_name, recipe_id)
+                                                    for recipe_name, recipe_id in recipe_name_ids ]
+            recipe_name_ids_content_string = ", ".join(recipe_name_ids_content_string_list)
+            content_string_list.append("recipes: {%s}" % recipe_name_ids_content_string)
+        yaml_content_string = "{%s}" % ", ".join(content_string_list)
```
haasken-hpe, line 259: I mentioned this to Mitch already in a slack message, but I'm wondering if we could use PyYAML to create the YAML string from a Python object instead, so we don't have to construct the YAML string manually. It's a little harder to follow than it needs to be.

scripts/operations/configuration/update_product_catalog_ims_ids.py
```diff
@@ -0,0 +1,156 @@
...
+            continue
+        if "id" not in imgrec_data:
+            LOGGER.debug("Product catalog entry for '%s' version '%s' contains %s '%s' with no "
+                         "'id' data: %s", product_name, product_version, images_or_recipes,
+                         imgrec_name, str(imgrec_data))
+            continue
+        imgrec_id = imgrec_data["id"]
+        if imgrec_id in id_change_map:
+            # This means that this ID was changed during the import
+            name_id_list.append((imgrec_name, id_change_map[imgrec_id]))
```
haasken-hpe, line 115: Generally speaking, I don't think this should happen given that an IMS export/import is going to change the IMS IDs of all recipes and images that were recorded in the product catalog. However, if there were any entries for images or recipes whoser IDs did not change, it looks like they'd be omitted from the `name_id_list` return value here. They would then be left out of the `YAML_CONTENT_STRING` passed to the `cray-product-catalog-update` container. I don't remember if `cray-product-catalog-update` merges the dictionaries under the `images` and `recipes` keys. I think older versions certainly did not merge and just replaced the values completely, but newer versions of `cray-product-catalog-update` may handle this better now. It might be worth double-checking, just to be sure we aren't going to drop information from the cray-product-catalog.

scripts/operations/system_recovery/ims-import-export.py
```diff
@@ -32,8 +32,15 @@
 import tempfile
 import uuid
 from os import path
+from typing import Dict
 from urllib.parse import urlparse
 
+# Ugly temporary hack to get the product_catalog module from a different path.
+# This will go away as the helper modules make their way into libcsm
+sys.path.insert(1, os.path.dirname(os.path.abspath(__file__))+"/../configuration")
```
haasken-hpe, line 40: Better to use `os.path.join`?

haasken-hpe (approved):
Minor comments. The edge case about possibly dropping information from the product catalog if IMS image/recipe IDs did not change may not even be a concern if we expect every single IMS ID from the product catalog to have changed.

scripts/operations/configuration/python_lib/product_catalog.py
mharding-hpe: Fixed.

scripts/operations/system_recovery/ims-import-export.py
mharding-hpe: Fixed.

scripts/operations/configuration/update_product_catalog_ims_ids.py
mharding-hpe: For the 1.3 backport I made a change so that it does the merging on my end, thus making it unimportant whether or not the updater handles it.

PR #3582, opened 2023-04-04
CASMINST-6156 instruct user to stop before rebuilding m001 in IUF documentation

leliasen-hpe:
# Description
Instruct user to stop before rebuilding m001 in IUF documentation. It was unclear how many steps in a specific document should be followed, it is to clarify where the user should stop.

# Checklist

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -47,6 +47,7 @@ StatefulSets
 1TiB
 2-chassis
 2.0.0.L
+200Gbps
 200GBASE-CR4
 200GBASE-DR4
 200GBASE-SR4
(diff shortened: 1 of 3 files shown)


dzou-hpe (approved):

deantroe:
I think we should put something directly into Stage 2 to remind the reader when to return to the IUF instructions as it's easy to lose track of when you are supposed to stop once you hop to another page.  We've started adding statements like this in a few other places due to SSI feedback.  For example, in CASMINST-6147 I am working on something similar with text like this:
```
If this step was executed as a result of the [`management-nodes-rollout` with CSM upgrade](#operations/iuf/workflows/management_rollout.md#management-nodes-rollout-with-csm-upgrade)
instructions, return to that procedure and continue with the next step. Otherwise, proceed to the next step.
```

...and this:
```
If performing an upgrade of CSM and additional HPE Cray EX software products using the IUF,
return to the [Upgrade CSM and additional products with IUF](../operations/iuf/workflows/upgrade_csm_and_additional_products_with_iuf.md)
procedure. Otherwise, if performing an upgrade of only CSM, proceed to the next step.
```

bklei (approved):

deantroe (approved):

PR #3639, opened 2023-04-13
CASMCMS-8533: Create and document CFS import tool

mharding-hpe:
# Description

Create scripts to import CFS data generated by the CFS export tool.
Document same.

I tested the new tools on wasp and verified that it does not import things that it should not, and that it does import those things that it should import.

# Checklist


rbak-hpe (approved):

rustydb (approved):

rustydb:
This JIRA is not a release critical, so I suspect there were other reasons to push it into docs-csm instead of libcsm. Since I suspect this code will still move to libcsm, I'll skip the review since there's numerous things that stick out to me but won't matter once this is refactored.

PR #3733, opened 2023-04-28
ALT-307 Add WLM workarounds for Alice

dgloe-hpe:
# Description

Document the WLM workarounds needed for Alice IUF installation.

# Checklist

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -52,5 +52,20 @@ For a list of all features with an announced removal target, see [Removals](intr
 ## Known issues
 
 * UAIs use a default route that sends outbound packets over the CMN, this will be addressed in a future release so that the default route uses the CAN/CHN.
+* The Slurm installer released in CPE 23.03 (`cpe-slurm-23.03-sles15-1.2.10.tar.gz`) has an issue that causes failures when installed with the IUF. To work around the issue, run the following commands before the IUF `process-media` stage:
+
+  ```bash
+  tar -xf cpe-slurm-23.03-sles15-1.2.10.tar.gz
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -52,5 +52,20 @@ For a list of all features with an announced removal target, see [Removals](intr
 ## Known issues
 
 * UAIs use a default route that sends outbound packets over the CMN, this will be addressed in a future release so that the default route uses the CAN/CHN.
+* The Slurm installer released in CPE 23.03 (`cpe-slurm-23.03-sles15-1.2.10.tar.gz`) has an issue that causes failures when installed with the IUF. To work around the issue, run the following commands before the IUF `process-media` stage:
```
JohnNavitskyHPE, line 55: This looks good, but perhaps we could add a note indicating what to do if they fail to do this prior to installation.  Something like.

"If Slurm 23.03 is installed without the workaround in place, apply the workaround and re-install it with the '--force' option"

deantroe (approved):
This looks good to me, but in general I am confused about the situation with CPE in Alice.  It's my understanding that the 23.03 CPE docs don't have any IUF content in them at all (it's in 23.05).  If that's the case, I don't see how an admin will known how to install CPE 23.03 with IUF without instructions.  

Am I misunderstanding the docs situation?

dgloe-hpe:
> This looks good to me, but in general I am confused about the situation with CPE in Alice. It's my understanding that the 23.03 CPE docs don't have any IUF content in them at all (it's in 23.05). If that's the case, I don't see how an admin will known how to install CPE 23.03 with IUF without instructions.
> 
> Am I misunderstanding the docs situation?

Yes, you're right. There's no mention of the IUF in the CPE 23.03 admin guide. The customer will be on their own if they want to install 23.03 with the IUF. But I don't think we have any better alternatives.

PR #3819, opened 2023-05-24
SCICD-732: Workflow Output

larry-tordsen:
Reflect the addition of script_stdout to the workflow output.


larry-tordsen:
This doesn't look quite right, so I'm closing it for now.

PR #3820, opened 2023-05-24
CASMPET-6348 automated storage node upgrade

leliasen-hpe:
# Description
Automated storage node upgrade using argo. This has been tested on Drax.

# Checklist


bklei (approved):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -787,8 +787,8 @@ if [[ ${state_recorded} == "0" && $(hostname) == "${PRIMARY_NODE}" ]]; then
 
     # As boot parameters are added or removed, update these arrays.
     # NOTE: bootparameters_to_delete should contain keys only, nothing should have "=<value>" appended to it.
-    bootparameters_to_set=( "split_lock_detect=off" "psi=1" "rd.live.squashimg=rootfs" "rd.live.overlay.thin=0" )
-    bootparameters_to_delete=( "rd.live.squashimg" "rd.live.overlay.thin" )
+    bootparameters_to_set=( "split_lock_detect=off" "psi=1" "rd.live.squashimg=rootfs" "rd.live.overlay.thin=0" "rd.live.dir=${CSM_RELEASE}" "rd.live.overlay.reset=1")
```
dzou-hpe, line 790: does this get set during fresh install as well? i am thinking a scenario where it was fresh installed and someone try to execute rebuild. in that case, this prereq will not be run

dzou-hpe (approved):

dzou-hpe (approved):

upgrade/scripts/upgrade/prerequisites.sh
leliasen-hpe: I don't think this would be needed during a rebuild. This is for the case where we are booting node into new images, which happens during an upgrade and not a rebuild. @rustydb is this correct?

rustydb: @dzou-hpe only `rd.live.dir` is set during a fresh install, `rd.live.overlay.reset` is only set during an upgrade to ensure a new overlay is created for storage nodes. `rd.live.overlay.reset` isn't needed where `metal.no-wipe=0` is set.

leliasen-hpe: I have changed prerequisites. Originally, this PR had set both `rd.live.dir` and `rd.live.overlay.reset=1` in prerequisites. Prerequisites now sets `rd.live.dir` for all NCNs. I removed `rd.live.overlay.reset` from prerequisites. This is now run as part of the storage node upgrade. It could cause an issue if left in prerequisites since it would set `rd.live.overlay.reset=1` on all NCNs and worker nodes do not remove this setting so it may be problematic when powering on/off worker nodes. 
Storage nodes will set `rd.live.overlay.reset=1` since they are upgraded with `metal.no-wipe=1` and it will set `rd.live.overlay.reset=0` after the storage node upgrade.

rustydb (approved):

PR #3825, opened 2023-05-25
CASMINST-6403 Avoid creating new VMs for each Jenkinsfile matrix axis

mtupitsyn:
# Description

Reduce amount of Google Compute VMs needed to build docs-csm RPMs down to 1. Currently, it takes 3 VMs for each build. It takes a lot of time and resources to spin up all 3 VMs and do a fresh git clone onto each of them. This change uses docker container with mounted cone directory instead of 2 VMs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ rpm_package_source:
 	tar --transform 'flags=r;s,^,/${NAME}-${VERSION}/,' -cvjf $(SOURCE_PATH) $(SPEC_FILE) --exclude .git --exclude "*.spec" --exclude dist .
 
 rpm_build_source:
-	rpmbuild -ts $(SOURCE_PATH) --define "_topdir $(BUILD_DIR)"
+	rpmbuild -bs $(BUILD_DIR)/SPECS/$(SPEC_FILE) --define "_topdir $(BUILD_DIR)"
 
 rpm_build:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -45,7 +45,7 @@ prepare:
 	cp $(SPEC_FILE) $(BUILD_DIR)/SPECS/
 
 rpm_package_source:
-	tar --transform 'flags=r;s,^,/${NAME}-${VERSION}/,' --exclude .git --exclude dist -cvjf $(SOURCE_PATH) .
+	tar --transform 'flags=r;s,^,/${NAME}-${VERSION}/,' -cvjf $(SOURCE_PATH) $(SPEC_FILE) --exclude .git --exclude "*.spec" --exclude dist .
```
mtupitsyn, line 48: This ensures that exactly one `*.spec` file is packaged with RPM source. If there more then 1 such file, `rpmbuild` will complain about multiple spec files. This change ensure we package the same spec file which was also copied to `(BUILD_DIR)/SPECS/`.

rustydb: Makes sense.

rustydb (approved):

rustydb (approved):

phalseth-hpe (approved):

PR #3845, opened 2023-06-01
CASMPET-6612 Improve spire-postgres health check

kburns-hpe:
This check was neither idempotent, nor did it wait long enough. This updates the rolling of the spire-postgres pods to be during the spire chart upgrade and moves the check to be just before validation. This should both reduce the time it takes to get through this piece as well as increasing the likihood that it will be completed before the validation check times out.

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -515,38 +515,13 @@ if [[ ${state_recorded} == "0" && $(hostname) == "${PRIMARY_NODE}" ]]; then
         "${locOfScript}/util/upgrade-spire.sh"
 
     } >> "${LOG_FILE}" 2>&1
+    
+    echo "Rolling spire-postgres statefulset"
+    kubectl rollout restart -n spire statefulset spire-postgres
     record_state "${state_name}" "$(hostname)" | tee -a "${LOG_FILE}"
 else
     echo "====> ${state_name} has been completed" | tee -a "${LOG_FILE}"
 fi
-
-state_name="ROLL_SPIRE_POSTGRES"
-state_recorded=$(is_state_recorded "${state_name}" "$(hostname)")
-if [[ ${state_recorded} == "0" ]]; then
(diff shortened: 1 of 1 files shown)


upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -1145,6 +1120,34 @@ else
...
+    echo "Waiting for postgres operator to consider spire-postgres healthy (${i}/${max_checks})"
+    sleep ${wait_time}
+    i=$(( i+1 ))
+  done
+    
+  if [[ ${healthy} == "false" ]]; then
+    echo "ERROR: spire-postgres is not healthy after $(( max_checks * wait_time )) seconds"
+    exit 1
+  fi
+    record_state "${state_name}" "$(hostname)" | tee -a "${LOG_FILE}"
```
johren-hpe, line 1146: Indentation is off on this line.  Should like up with the line before it.

johren-hpe (changes requested):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -1145,6 +1120,34 @@ else
...
+else
+    echo "====> ${state_name} has been completed" | tee -a "${LOG_FILE}"
```
johren-hpe, line 1148: Two many spaces in this indentation.    The rest of this stage uses 2-space indentation.

kburns-hpe: I ran shfmt on it. I'm hitting too many issues trying to do this with my tools turned off.

kburns-hpe: I ran shfmt on it. I'm hitting too many issues trying to do this with my tools turned off.

upgrade/scripts/upgrade/prerequisites.sh
kburns-hpe: I ran shfmt on it. I'm hitting too many issues trying to do this with my tools turned off.

rustydb (approved):

PR #3846, opened 2023-06-01
CASMPET-6613 Add shfmt

kburns-hpe:
This adds running shellcheck and shfmt to the build pipeline. This will allow for a more consistent formatting among teams and reduce the need for reviewers to check for these issues.


mtupitsyn (approved):

kburns-hpe:
Looks like shellcheck was already in there. Removing it in favor of running shfmt in a similar way.

rustydb:
Discussing with libcsm/docs-csm ppl, this may need to merge with the shellcheck workflow and/or align with what we're doing in libcsm (since docs-csm workflows are a little stale). Will get back to you asap, and get this in.

rustydb (changes requested):
Please merge this with the existing shellcheck workflow, then we don't have to manage yet another workflow in the branch protection settings.

rustydb (approved):

PR #3847, opened 2023-06-01
CASMPET-6516

cdelatte-hpe:
# Description
This addresses insecure registry running on the utility storage nodes. It moves those images to Nexus and disables the service running the registry.

# Checklist

# Noted
- [x] Confirmed on Fanta after 1.3 install

### DIFF
--- a/scripts/upload_ceph_images_to_nexus.sh
+++ b/scripts/upload_ceph_images_to_nexus.sh
@@ -23,14 +23,28 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 
-m001_ip=$(host ncn-m001 | awk '{ print $NF }')
-ssh-keygen -R ncn-m001 -f ~/.ssh/known_hosts > /dev/null 2>&1
-ssh-keygen -R "${m001_ip}" -f ~/.ssh/known_hosts > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


med-hpe:
systemctl_force is typoed in line 320 (so line 317 reports never used but fixing that should fix both of these)

med-hpe:
I'm still taking my first look at this patchset

scripts/upload_ceph_images_to_nexus.sh
```diff
@@ -312,11 +312,89 @@ function exit_maintenance_mode() {
...
+       if ! ssh ${storage_node} ${ssh_options} "systemctl is-enabled registry.container.service"; then
+         echo "Docker registry service on ${storage_node} has been disabled"
+       fi
+    fi
+  done
+}
+
+function fix_registries_conf() {
+
+  for storage_node in $(ceph orch host ls -f json |jq -r '.[].hostname'); do
```
med-hpe, line 330: This creates a here file each time through. Probably simpler to just create it once and then loop (by moving for down a tad bit). We can also probably use something smoother than cat-ing that file into an ssh (like scp, rsync, etc). Probably be easier to follow that way.
Where is this running from (I presume one of the storage nodes and then it updates that same said storage node. Might be easier to skip the current node.)

rustydb (approved):

PR #3857, opened 2023-06-07
CASMPET-6609 add script for upgrading and monitoring Ceph upgrade

leliasen-hpe:
# Description
Ceph-upgrade-tool can be used to start a ceph upgrade and to monitor the upgrade. It will print status updates until the Ceph upgrade completes or until there is an upgrade error in which case it the script will exit.
This tool was incorporated into the storage-node upgrade workflow so now the ceph upgrade will be automated as part of the storage node upgrade.

This script was tested on Mug, beau, and dorian. I also ran the new argo stage 'upgrade-ceph' on Mug and validated that the script functions when called by an argo workflow.

# Checklist


bklei (approved):

workflows/templates/storage.ceph-upgrade.yaml
```diff
@@ -0,0 +1,58 @@
...
+                    for storage_node in $(ceph orch host ls -f json |jq -r '.[].hostname'); do
+                      if [[ -z $(ssh ${storage_node} ${ssh_options} 'cat /etc/os-release' | grep 'VERSION="15-SP4"') ]]; then
+                        echo "Not all storage nodes have been upgraded and are running SP4."
+                        echo "Waiting to upgrade Ceph until all storage nodes have been upgraded."
+                        exit 0
+                      fi
+                    done
+                    export TERM=linux
+                    # setting PYTHONUNBUFFERED forces python to print output
+                    export PYTHONUNBUFFERED=0
```
med-hpe, line 57: While this line works as is it may be more clear to do something like:
export PYTHONUNBUFFERED=true
or
export PYTHONUNBUFFERED=1
as setting it to 0 is not entirely obvious.

PR #3880, opened 2023-06-12
CASMINST-6470: Update chart redeploy procedure

mharding-hpe:
# Description
In a number of places in the docs we have administrators redeploy charts, for a variety of reasons. These are not done in a uniform way. Many of them assume that the loftsman manifest from the CSM install/upgrade will exist unchanged in the loftsman Kubernetes configmap, while others overwrite it, in some cases overwriting it with most of the charts removed.

I raised this issue in a Slack DM and there has been much discussion about what we should be doing. I am not an expert on this, but I do know we need to make some change to the docs to at least make sure we aren't overwriting things, or having customers regress product versions inadvertently.

This PR updates chart redeploy procedure to prevent overwriting existing manifests while still using correct versions and customizations. It is not meant to be the ideal eventual solution to the problem, It is just meant to get rid of the worst problems we currently face:
1. Procedures that overwrite previously-deployed manifests, including those from the CSM install/upgrade
2. Procedures which assume that the "correct" version for a chart is the one from that manifest, even though hotfixes may have been subsequently deployed
3. Generally inconsistent procedures

It does this by making one page which outlines how to redeploy a chart. The high level procedure it uses is:
1. Download the customizations from Kubernetes and (optionally) make changes.
2. Download the base CSM install/upgrade manifest for the chart in question
3. Make a copy of that manifest file that strips out all of the charts except for the one being redeployed, and which gives the manifest a unique name (so it won't overwrite the original when deployed)
4. Look to see if the most recent loftsman deployment of the chart (which may be more recent than the CSM upgrade/install) is using a different version. If so, update your copied file with that version number.*
  * I used a small shell script for this, because it's tedious work. I originally wrote the shell script when toying with ways to handle this for CSM hotfix deploys. We ended up not using it there, but I had tested it pretty well for that purpose. It seems better and less error-prone than having the admin have to do that stuff manually.
5. Apply the customizations to your file.
6. Redeploy the chart
7. If you made customizations changes, then upload them back into K8s

The plan is for other pages to reference this procedure when needing to redeploy a chart.

**PLEASE BEAR IN MIND that this is NOT meant to be the eventual solution. This is just meant to be something that isn't going to break customer systems the way our current procedures risk doing.**

Backports:
- [1.4](https://github.com/Cray-HPE/docs-csm/pull/3897)
- [1.3](https://github.com/Cray-HPE/docs-csm/pull/3898)
- [1.2](https://github.com/Cray-HPE/docs-csm/pull/3899)
- [1.0](https://github.com/Cray-HPE/docs-csm/pull/3900)

# Checklist

### DIFF
--- a/operations/CSM_product_management/Post_Install_Customizations.md
+++ b/operations/CSM_product_management/Post_Install_Customizations.md
@@ -132,22 +132,10 @@ Use Grafana to investigate and analyze CPU throttling and memory usage.
 
 ### Prerequisites
 
-In order to apply post-install customizations to a system, the affected Helm chart must exist on the system so that the same chart version can be redeployed with the desired customizations.
-
-This example unpacks the `csm-1.0.0` tarball under `/root` and lists the Helm charts that are now on the system.
-Set `PATH_TO_RELEASE` to the release directory where the `helm` directory exists.
-`PATH_TO_RELEASE` will be used below when deploying a customization.
-
-(`ncn-mw#`) These unpacked files can be safely removed after the customizations are deployed through `loftsman ship` in the examples below.
-
-```bash
-## This example assumes the csm-1.0.0 release is currently running and the csm-1.0.0.tar.gz has been pulled down under /root
-cd /root
-tar -xzf csm-1.0.0.tar.gz
-rm csm-1.0.0.tar.gz
-PATH_TO_RELEASE=/root/csm-1.0.0
-ls $PATH_TO_RELEASE/helm
-```
+Most of these procedures instruct the administrator to perform the [Redeploying a Chart](Redeploying_a_Chart.md)
+procedure for a specific chart. In these cases, the section on this page provides the administrator with the
+information necessary in order to carry out that procedure. It is recommended to keep both pages open
+in different browser windows for easy reference.
 
 ### Prometheus pod is `OOMKilled` or CPU throttled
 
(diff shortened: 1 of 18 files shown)


operations/CSM_product_management/Post_Install_Customizations.md
```diff
@@ -291,76 +238,41 @@ This example is based on what was needed for a system with 4000 compute nodes.
...
+    ```bash
+    rpname=<put resource path name from table here>
+    ```
 
-1. (`ncn-mw#`) Generate the manifest that will be used to redeploy the chart with the modified resources.
+1. (`ncn-mw#`) Edit the customizations by adding or updating `spec.kubernetes.services.${rpname}.cray-postgresql.sqlCluster.resources`.
 
    ```bash
-   manifestgen -c customizations.yaml -i sysmgmt.yaml -o manifest.yaml
+   yq write -i customizations.yaml "spec.kubernetes.services.${rpname}.cray-postgresql.sqlCluster.resources.requests.cpu" --style=double '4'
```
kburns-hpe, line 266: I find this section confusing. While it tells you to follow the redeploy_a_chart.md instructions before, it's not close enough to this section. This leads me to believe people will modify the customizations.yaml file here and wonder why that didn't update their resources.

operations/CSM_product_management/Post_Install_Customizations.md
```diff
@@ -291,76 +238,41 @@ This example is based on what was needed for a system with 4000 compute nodes.
...
+   yq write -i customizations.yaml "spec.kubernetes.services.${rpname}.cray-postgresql.sqlCluster.resources.requests.memory" '4Gi'
```
kburns-hpe, line 267: Requests and limits for memory resources should be the same. Having differences between these two, especially 12Gb swings for a single postgres cluster, can cause unexpected OOMs that may be difficult to track back to this being the problem.

operations/CSM_product_management/Post_Install_Customizations.md
```diff
@@ -291,76 +238,41 @@ This example is based on what was needed for a system with 4000 compute nodes.
...
+   yq write -i customizations.yaml "spec.kubernetes.services.${rpname}.cray-postgresql.sqlCluster.resources.limits.cpu" --style=double '8'
```
kburns-hpe, line 268: I'll also note that it's common practice now to not add a cpu limit on pods. Since we do this everywhere, I'm unsure if we should change anything here without having a larger discussion on right-sizing pods.

operations/CSM_product_management/Post_Install_Customizations.md
```diff
@@ -374,90 +286,59 @@ Refer to the note at the end of this section for more details.
...
+    ```
 
 ### Scale `cray-bss` service
 
 Scale the replica count associated with the `cray-bss` service in the `services` namespace.
 This example is based on what was needed for a system with 4000 compute nodes.
 Trial and error may be needed to determine what is best for a given system at scale.
 
-1. (`ncn-mw#`) Get the current cached customizations.
+Follow the [Redeploying a Chart](Redeploying_a_Chart.md) procedure to redeploy the
```
kburns-hpe, line 333: This also feels like it's too early vs the steps provided.

operations/CSM_product_management/Redeploying_a_Chart.md
```diff
@@ -0,0 +1,214 @@
+# Redeploying a Chart
```
kburns-hpe, line 1: This entire document feels like it should be boiled down to 1 or 2 scripts, whose use can be referenced somewhere else.

operations/CSM_product_management/Post_Install_Customizations.md
mharding-hpe: I didn't change any of the content of the values being changed. This PR is just to change the actual process. Unless the non-proecedure-related feedback is very limited, I'll probably open a separate ticket for those fixes.

operations/CSM_product_management/Post_Install_Customizations.md
mharding-hpe: Isn't the direction to follow that procedure directly only a couple of paragraphs higher up?
After all, if they don't follow that procedure, they won't have the customizations.yaml there to be edited, or the manifest file to be deployed.

I can see where it could be confusing at first glance, but how could someone modify the customizations file if they weren't following that procedure?

operations/CSM_product_management/Post_Install_Customizations.md
mharding-hpe: Definitely a separate discussion. This PR is only for fixing the procedure we follow when telling customers to redeploy charts. That is already going to be extensive enough without trying to pull in other stuff. :)

operations/CSM_product_management/Post_Install_Customizations.md
mharding-hpe: This is why it tells you to keep both pages open in the browser. You are following that procedure, but you need information from this page in order to do it. The alternatives to having it be in this format are:

1. Don't document the procedure in a central place, and instead repeat that procedure in all places where we want to charts to be redeployed. The downside of this approach is that we do this in a non-trivial number of places in the docs, and historically we have not done a good job of making sure that changes to a procedure get propagated to every place where that procedure appears in the docs. 

2. As you noted in another comment, automate this stuff. I'll respond to that comment in its own thread.

operations/CSM_product_management/Redeploying_a_Chart.md
mharding-hpe: I agree. The problem is:
1) Scripts require far more testing than manual instructions, since you have to check for things that a human would obviously realize are problems but that a script would not.
2) It will take me longer just to write the scripts.
3) We want to avoid adding scripts to docs-csm and instead put them in libcsm, but for that repo we are trying not to add any scripts which don't have accompanying unit tests. That makes it more work again.

I think automation is the eventual solution.

I cannot repeat enough that the goal here is to provide a bandaid that fixes the problems we have seen with the existing procedures -- namely, that they overwrite the manifests in Kubernetes, and that they do not track cases where version numbers have been updated due to hotfixes. It is a secondary step to provide a more elegant and proper solution to this. I don't have the time to dedicate to that right now. That said, if you do happen to have time to tackle it, I am happy to assign this over to you. :)

operations/CSM_product_management/Post_Install_Customizations.md
mharding-hpe: I will try to add some language and markup that make it much clearer that you need to be following the general procedure and only use the other sections when you get to those parts.

kburns-hpe (approved):

mharding-hpe:
Working on backports

nrockershousen (approved):

rustydb (approved):

PR #3886, opened 2023-06-14
CASMINST-6467 run verify images from within each management-node workflow

leliasen-hpe:
# Description
For IUF stage management-nodes-rollout, each management node type has its own workflow that can be run independently. The images and CFS config should be verified as part of each individual workflow. Sometimes, only worker nodes will need to be rebuilt and if verify images is always checking for storage node images as well, the step may fail even though it is not blocker the worker nodes being rolled out. For this reason, each workflow should verify the images for that node type only.

I tested basic functionality on Drax and saw that the verify-ncn-images-configs template works correctly. I did notice a logging bug and opened [CASMINST-6485](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6485) to address it.

# Checklist


workflows/iuf/stages.yaml
```diff
@@ -118,15 +118,7 @@ stages:
...
-      - name: info-to-read
-        static-parameters: {} # any parameters that will be supplied statically to this operation.
-      - name: verify-images-and-configuration
-        static-parameters: {} # any parameters that will be supplied statically to this operation.
-      - name: management-storage-nodes-rollout
-        static-parameters: {} # any parameters that will be supplied statically to this operation.
-      - name: management-two-master-nodes-rollout
-        static-parameters: {} # any parameters that will be supplied statically to this operation.
-      - name: management-worker-nodes-rollout
+      - name: management-nodes-rollout
```
sravani-sanigepalli, line 121: With the changes of refactoring, there is no operation `management-nodes-rollout`. All the refactored operations should be called individually.

workflows/iuf/operations/management-nodes-rollout/management-storage-nodes-rollout.yaml
```diff
@@ -64,7 +64,7 @@ spec:
             templateRef:
               name: workflow-template-record-time-template
               template: record-time-template
-          - name: verify-storage-images-and-configuration
+          - name: INFO-to-read
```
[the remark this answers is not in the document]
leliasen-hpe: I think this is a good idea generally, however, INFO-to-read is unique for each stage. It provides information about what that stage is doing. It would add more complication to try to have 1 common template with different output based on node type. In my opinion, it should be left as is.

workflows/iuf/operations/management-nodes-rollout/verify-ncn-images-configs.yaml
```diff
@@ -0,0 +1,88 @@
+#
```
sravani-sanigepalli, line 1: `verify-ncn-images-configs.yaml` is now enhanced version of `verify-images-and-configuration.yaml` which is made more modular. With these changes, can `verify-images-and-configuration.yaml` be removed as well??
[https://github.com/Cray-HPE/docs-csm/blob/CASMINST-6467/workflows/iuf/operations/management-nodes-rollout/verify-images-and-configuration.yaml]

workflows/iuf/stages.yaml
leliasen-hpe: This is a good catch but this stage works differently than the others. The IUF backend handles getting the correct workflow for NCN type based on inputs to --limit-management-rollout parameter. The code that does this is [here](https://github.com/Cray-HPE/cray-nls/blob/0597b3fb49c966d260fac173b82bd224a3467af3/src/api/services/iuf/sessions_workflow_gen.go#L636).

workflows/iuf/operations/management-nodes-rollout/management-storage-nodes-rollout.yaml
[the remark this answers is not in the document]
leliasen-hpe: I misread your comment, yes the general template should be removed. I will remove it.

workflows/iuf/operations/management-nodes-rollout/verify-ncn-images-configs.yaml
leliasen-hpe: Yes, I will remove the `verify-images-and-configuration.yaml` file.

workflows/iuf/stages.yaml
sravani-sanigepalli: Ahh, This is new information to me. Good to know that. Thanks.

med-hpe (approved):
lgtm

PR #3918, opened 2023-06-27
CASMINST-6486 have update-bss and update-cfs use ssh-template

leliasen-hpe:
# Description
After a change to iuf-base-template that this the update bss and cfs step were using, these steps no longer worked. As this workflow is sometime run outside of IUF, it would be good for it to reference a different workflowTemplate. It now uses ssh-template and the steps run successfully. This was tested on Fanta during management-nodes-rollout.

# Checklist

### DIFF
--- a/workflows/ncn/worker/worker.drain.yaml
+++ b/workflows/ncn/worker/worker.drain.yaml
@@ -63,7 +63,7 @@ tasks:
             done
   - name: update-bss
     templateRef:
-      name: iuf-base-template
+      name: ssh-template
       template: shell-script
     arguments:
       parameters:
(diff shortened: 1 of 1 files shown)


atifsyedali (approved):

workflows/ncn/worker/worker.drain.yaml
```diff
@@ -63,7 +63,7 @@ tasks:
             done
   - name: update-bss
     templateRef:
-      name: iuf-base-template
+      name: ssh-template
```
zodiac1214, line 66: can we just use `kubectl-and-curl-template`? i am guessing the problem here is that iuf-base-template no long mount /host_usr_bin ? i think `kubectl-and-curl-template` still does so we can just use that (it also exports TOKEN automatically)

leliasen-hpe: I don't remember the kubectl and curl template being able to run 'cray' commands which I think is why I didn't use it initially. Is it able to run 'cray' 
commands?

dzou-hpe: i see. that's the reason why Atif removed that usr_host_bin lol

dzou-hpe (approved):

rustydb (approved):

PR #3955, opened 2023-07-05
CASMINST-6528 Adds script to dynamically update tag references to containers in scr…

denniswalker:
…ipts and argo files.

# Description

# Checklist


scripts/update_tags.sh
```diff
@@ -0,0 +1,70 @@
...
+
+function get_latest_tag_for_image() {
+  THIS_REPO=$(get_repository_name_for_image $THIS_IMAGE)
+  podman search $THIS_REPO --list-tags --format=json | jq -r '.[0].Tags | last'
+}
+
+LATEST_TAG=$(get_latest_tag_for_image $THIS_IMAGE)
+
+function get_filenames_referring_to_image() {
+  grep -RHl -e "${THIS_IMAGE}:[0-9]\.[0-9]\.[0-9]" .
```
johren-hpe, line 57: There are some images that have a "v" in the tag.  e.g. cray-tapms-operator:v0.0.11

scripts/update_tags.sh
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function get_latest_tag_for_image() {
+  THIS_REPO=$(get_repository_name_for_image $THIS_IMAGE)
+  podman search $THIS_REPO --list-tags --format=json | jq -r '.[0].Tags | last'
+}
+
+LATEST_TAG=$(get_latest_tag_for_image $THIS_IMAGE)
+
+function get_filenames_referring_to_image() {
```
johren-hpe, line 56: Will need to be careful here.   Depending on which directory is given as an argument, this may pick up the "scans" directory and end up modifying the snyk scan data.

```
ncn-m001:/etc/cray/upgrade/csm/csm-1.4.1/tarball/csm-1.4.1 # grep -RHl -e "${THIS_IMAGE}:[0-9]\.[0-9]\.[0-9]" .
./scans/docker/artifactory.algol60.net/csm-docker/stable/cray-ims-load-artifacts:2.2.1/snyk.json
./docs/operations/package_repository_management/Troubleshoot_Nexus.md
./docs/workflows/iuf/operations/ims-upload.yaml
```

It also ends up modifying troubleshooting example output in this case.

scripts/update_tags.sh
```diff
@@ -0,0 +1,70 @@
...
+  grep -RHl -e "${THIS_IMAGE}:[0-9]\.[0-9]\.[0-9]" .
+}
+
+function update_tags_in_file() {
+  THIS_FILE=$1
+  echo "Updating tag of ${THIS_IMAGE} in ${THIS_FILE} to ${LATEST_TAG}."
+  sed -i -e "s/${THIS_IMAGE}:[0-9]\.[0-9]\.[0-9].*/${THIS_IMAGE}:${LATEST_TAG}/g" $THIS_FILE
```
johren-hpe, line 63: This also doesn't work for images that have "v" in the tag name.

scripts/update_tags.sh
```diff
@@ -0,0 +1,70 @@
...
+}
+
+for FILE in $(get_filenames_referring_to_image); do
+  update_tags_in_file $FILE
+done
+
+cd $OLD_PWD
```
johren-hpe, line 70: OLD_PWD isn't being set anywhere.

johren-hpe (changes requested):

scripts/update_tags.sh
johren-hpe: We also have this tag:  etcd-operator:0.12.6-cray

atifsyedali:
About where to run this script...

So the script in docs-csm `workflows/scripts/upload-rebuild-templates.sh` is called from both upgrade and install:

- Upgrade: from `prerequisites.sh` — happens somewhere in the middle before all images and charts are updated
- Install: from `install/scripts/csm_services/steps/7.docs_csm_upload_rebuild_templates.yaml`  — happens at the end.

So theoretically, we add your image tag replacement script in `workflows/scripts/upload-rebuild-templates.sh` and have it be called from both install and upgrade.

For upgrade, the slight change we will have to make is to move the call to `workflows/scripts/upload-rebuild-templates.sh` towards the end of the `prerequisites.sh` script to first allow all helm/Docker updates first.

scripts/update_tags.sh
denniswalker: It's a built-in variable to bash. Ah, but I need to remove the underscore. https://techpiezo.com/linux/pwd-vs-oldpwd-bash-environment-variables/

scripts/update_tags.sh
```diff
@@ -0,0 +1,70 @@
...
+  exit 1
+fi
+
+function get_repository_name_for_image() {
+  podman search registry.local/$THIS_IMAGE --format=json | jq -r '.[0].Name'
+}
+
+function get_latest_tag_for_image() {
+  THIS_REPO=$(get_repository_name_for_image $THIS_IMAGE)
+  podman search $THIS_REPO --list-tags --format=json | jq -r '.[0].Tags | last'
```
denniswalker, line 51: Probably need a sort here just in case results aren't sorted from the API.

scripts/update_tags.sh
denniswalker: fixed

scripts/update_tags.sh
denniswalker: fixed

scripts/update_tags.sh
denniswalker: fixed

scripts/update_tags.sh
denniswalker: fixed

scripts/update_tags.sh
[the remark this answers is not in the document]
denniswalker: fixed

workflows/update_tags.sh
```diff
@@ -0,0 +1,77 @@
...
+  exit 1
+fi
+
+if [[ $(nslookup $REGISTRY >> /dev/null; echo $?) != 0 ]]; then
+  echo "INFO: Skipping update_tags.sh script given docs were installed outside of a CSM environment."
+  exit 0
+fi
+
+if [[ ! $(curl -s https://${REGISTRY}) ]]; then
+  echo "ERROR: Is this a CSM environment? Check that ${REGISTRY} is accessible over https before continuing."
```
rustydb, line 48: I think this will cause a misleading failure when docs-csm is installed on the PIT node.

rustydb (changes requested):

workflows/update_tags.sh
denniswalker: In conversation, before this, check for /etc/pit-release existence and exit with message that this doesn't run on PIT.

denniswalker: fixed

rustydb (approved):

PR #4000, opened 2023-07-13
Missing triple-backtick code block formatting

pawsey-kbuckley:
Been a while since I checked the "validty" of your Markdown against
a Jekyll rendering, so top marks for only introducing one place, where
a "closing" triple-backtick is missing, in all that time

FWIW, a bit further down in the same file, you do close a similar block.

As always, HTH,
Kevin

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -32,6 +32,10 @@ StatefulSets
 
 
 1.0.x
+1.2.x
+1.3.x
+1.4.x
+1.5.x
 1.6.x
 100GBASE-CR2
 100GBASE-CR4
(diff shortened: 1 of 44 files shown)


nrockershousen (approved):

pawsey-kbuckley:
So what's with the licence check here?
This PR add three backticks into one file: what licensing is needed for that activity?
More than happy to comply though: if I knew what I was needed to be in compliance with.

rustydb:
This PR closed because it was fixed in another PR.

PR #4071, opened 2023-07-28
CASMTRIAGE-5761 Remove strimzi from csm-upgrade

kburns-hpe:
This is leftover from an older upgrade and isn't needed for 1.5.


kburns-hpe:
This PR isn't auto updating. Going to close it and create a new one.

PR #4083, opened 2023-08-01
CASMINST-6583 Update `update_tags.sh`

denniswalker:
# Description

# Checklist


workflows/update_tags.sh
```diff
@@ -25,28 +25,37 @@
...
   echo "ERROR: This script must be located in the workflows directory to run."
   exit 1
 fi
 
-if [[ $(nslookup $REGISTRY >> /dev/null; echo $?) != 0 ]]; then
-  echo "INFO: Skipping update_tags.sh script given docs were installed outside of a CSM environment."
-  exit 0
+# Change registry URL target for PIT
+if [[ -f /etc/pit-release ]]; then
+  echo "INFO: Running on PIT. Updating registry URL to $REGISTRY_NAME."
```
johren-hpe: Whether we use `registry.local` or `pit.nmn:5000` doesn't really depend on where the script is running.   This can run on m002 or w003 when `registry.local` is not available and it can run on the pit when `pit.nmn:5000` is not available.  Instead, I would try `registry.local` and if that fails then try `pit.nmn:5000`.   And if that fails, then you print out the warning.

johren-hpe: Or you could skip `pit.nmn:5000` altogether if there is something that will be updating the tags later when CSM services are installed.

johren-hpe (changes requested):

workflows/update_tags.sh
```diff
@@ -25,28 +25,35 @@
...
-  echo "INFO: Skipping update_tags.sh script given docs were installed on PIT."
-  exit 0
-fi
+[[ ! $(curl -s ${THIS_REGISTRY_PROTOCOL}://${THIS_REGISTRY_NAME}) ]] && update_registry_url_to_pit
 
-if [[ ! $(curl -s https://${REGISTRY}) ]]; then
-  echo "ERROR: Is this a CSM environment? Check that ${REGISTRY} is accessible over https before continuing."
-  exit 1
+if [[ ! $(curl -s ${THIS_REGISTRY_PROTOCOL}://${THIS_REGISTRY_NAME}) ]]; then
+  echo "WARNING: Unable to update workflow image tag references. ${THIS_REGISTRY_NAME} is not accessible."
```
johren-hpe: This may be confusing because it will always say "pit.nmn:5000 is not accessible".   That may lead one to think that they need to get PIT nexus up it will likely usually be registry.local that needs to be accessible.

johren-hpe (commented):
Requested changes made.  Just leaving regarding potentially confusing warning message.

workflows/update_tags.sh
denniswalker: done

workflows/update_tags.sh
denniswalker: Changed to "the registry is not accessible".

PR #4127, opened 2023-08-14
CASMPET-6747 upgrade ceph after storage nodes have SP5

leliasen-hpe:
# Description
After the storage node images are upgraded, Ceph is upgraded by the Argo workflow. The logic in the workflow check to see that all storage nodes are running the new os-version. This was previously checking for SP4, however, now we are upgrading to SP5 on the storage nodes so the check needs to look for SP5 on storage nodes.

# Checklist

### DIFF
--- a/workflows/templates/storage.ceph-upgrade.yaml
+++ b/workflows/templates/storage.ceph-upgrade.yaml
@@ -46,8 +46,8 @@ spec:
                   value: |
                     ssh_options="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
                     for storage_node in $(ceph orch host ls -f json |jq -r '.[].hostname'); do
-                      if [[ -z $(ssh ${storage_node} ${ssh_options} 'cat /etc/os-release' | grep 'VERSION="15-SP4"') ]]; then
-                        echo "Not all storage nodes have been upgraded and are running SP4."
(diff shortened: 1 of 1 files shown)


workflows/templates/storage.ceph-upgrade.yaml
```diff
@@ -46,8 +46,8 @@ spec:
                   value: |
                     ssh_options="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
                     for storage_node in $(ceph orch host ls -f json |jq -r '.[].hostname'); do
-                      if [[ -z $(ssh ${storage_node} ${ssh_options} 'cat /etc/os-release' | grep 'VERSION="15-SP4"') ]]; then
-                        echo "Not all storage nodes have been upgraded and are running SP4."
+                      if [[ -z $(ssh ${storage_node} ${ssh_options} 'cat /etc/os-release' | grep 'VERSION="15-SP5"') ]]; then
+                        echo "Not all storage nodes have been upgraded and are running SP5."
                         echo "Waiting to upgrade Ceph until all storage nodes have been upgraded."
```
rustydb, line 51: This won't work on subsequent upgrades, at least it won't catch anything when going from SP5 -> SP5. Is this strictly just to keep the service packs the same between the ceph nodes?

PR #4129, opened 2023-08-14
CASM-4505 release-16

atifsyedali:
# Description

The update_tags.sh script was introduced in [CASMINST-6528](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6528) to mitigate the possibility of not having the images declared in latest doc on the system, and to fall back to whatever is available.

The problem is that the images that are chosen on the system are not based on semantic versioning, rather on a textual sort. This is incorrect as it ends up selecting previous versions. E.g. between 0.1.10 and 0.1.3, it will select 0.1.3 which is incorrect.

The solution is to add semantic version comparisons and sort before selecting the latest image.

Tested with different types of version tags, including ones with textual prefixes and suffixes in the version tag.

# Checklist


atifsyedali:
bad PR. Going to recreate

PR #4164, opened 2023-08-22
CASMTRIAGE-5925 use default value when myenv doesn't exist

dzou-hpe:
# Description

# Checklist

### DIFF
--- a/workflows/ncn/worker/worker.wipe-and-reboot.yaml
+++ b/workflows/ncn/worker/worker.wipe-and-reboot.yaml
@@ -140,7 +140,7 @@ tasks:
             
             # wait for boot
             counter=0
-            source /etc/cray/upgrade/csm/myenv
+            source /etc/cray/upgrade/csm/myenv || true
             if [[ -z $REBOOT_TIMEOUT_IN_SECONDS ]];then
                 REBOOT_TIMEOUT_IN_SECONDS=600
             fi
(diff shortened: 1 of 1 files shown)


dzou-hpe:
note: no need to backport to 1.5 and plus, we have implemented it in a different way there

bo-quan (approved):

leliasen-hpe (approved):

workflows/ncn/worker/worker.wipe-and-reboot.yaml
```diff
@@ -140,7 +140,7 @@ tasks:
             
             # wait for boot
             counter=0
-            source /etc/cray/upgrade/csm/myenv
+            source /etc/cray/upgrade/csm/myenv || true
```
rustydb: Will this mask nefarious things in `myenv`? Would it be better to do the following?

```bash
[ -f /etc/cray/upgrade/csm/myenv ] && source /etc/cray/upgrade/csm/myenv
```

mharding-hpe: I think your modified version will fail if the code runs with `set -e` (which I assume it does, because of the `|| true`).

I think your change makes sense, but it would need to be modified to something like:

```suggestion
            [[ ! -f /etc/cray/upgrade/csm/myenv ]] || source /etc/cray/upgrade/csm/myenv
```

rustydb (approved):

PR #4299, opened 2023-09-25
CASMINST-6649-release-1.5 enable smartmon on storage nodes before upgrading first storage node

leliasen-hpe:
# Description

When storage nodes are upgraded to CSM 1.5, smartmon is enabled on storage nodes. Node-exporter daemons on non-upgraded storage nodes get into a bad state because the non-upgraded storage nodes do not have the neccessary rpms installed. This change installs smartmon and the needed rpms on storage nodes before they are upgraded so node-exporter does not error.

# Checklist


bklei (approved):

ndavidson-hpe (approved):

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,92 @@
...
+ssh_options="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+
+function get_smartmon_url() {
+  local repos
+
+  if ! curl -sSfk https://packages.local/service/rest/v1/repositories >&/dev/null; then
+    echo "***"
+    echo "WARNING: Unable to contact Nexus! This must be addressed before proceeding."
+    echo "***"
+    return
```
rustydb: Should return a non-zero.

I think `return` by itself returns the last return code, which would be `0` from the previous `echo` call.

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,92 @@
...
+    {
+      token="$(curl -sSk "$url&continuationToken=${token}" | tee /dev/fd/3 | jq -r '.continuationToken // null')"
+    } 3>&1
+
+    if test -z $token; then
+      echo "ERROR on line $LINENO: unable to retreive continuation token, exiting"
+      exit 1
+    fi
+  done
+}
```
rustydb: `s/exit/return/` function should return instead of exiting the script, whatever is calling `paginate` should exit if its return is non-zero.

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,92 @@
...
+
+get_smartmon_url
+
+echo "Using ${smartmon_url} to install rpm on storage nodes..."
+
+for storage_node in $(ceph orch host ls -f json | jq -r '.[].hostname'); do
```
rustydb: This will safely create an array.
```suggestion
if IFS=$'\n' read -rd '' -a ORCH_HOSTS; then
        :
fi <<< "$(ceph orch host ls -f json | jq -r '.[].hostname')"
for storage_node in "${ORCH_HOSTS[@]}"; do
```

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,92 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+
```
rustydb, line 25: Can this work with `set -euo pipefail`?

rustydb (changes requested):

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,92 @@
...
+    return
+  fi
+
+  # get a list of repos that start with "csm-noos"
+  repos=$(curl -sSfk https://packages.local/service/rest/v1/repositories | jq -r '.[] | .["name"]' | grep ^csm-noos | tr '\n' ' ')
+
+  echo "Will look for smart-mon rpm in the following repos: $repos"
+
+  # search through the csm-noos repos looking for our package
+  for repo in $repos; do
```
rustydb: This will safely create an array, and let you use double-quotes.

```suggestion
  # get a list of repos that start with "csm-noos"
  if IFS=$'\n' read -rd '' -a repos; then
        :
  fi <<< "$(curl -sSfk https://packages.local/service/rest/v1/repositories | jq -r '.[] | .["name"]' | grep ^csm-noos | tr '\n' ' ')" 

  echo "Will look for smart-mon rpm in the following repos: ${repos[@]}"

  # search through the csm-noos repos looking for our package
  for repo in "${repos[@]}"; do
```

rustydb (changes requested):

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
leliasen-hpe: fixed

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "ERROR on line $LINENO: unable to retreive continuation token, exiting"
+    return 1
+  fi
+
+  until [[ $token == "null" ]]; do
+    {
+      token="$(curl -sSk "$url&continuationToken=${token}" | tee /dev/fd/3 | jq -r '.continuationToken // null')"
+    } 3>&1
+
+    if test -z $token; then
```
rustydb, line 91:
```suggestion
    if [ -z "$token" ]; then
```

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "***"
+    return 1
+  fi
+
+  # get a list of repos that start with "csm-noos"
+  if IFS=$'\n' read -rd '' -a repos; then
+        :
+  fi <<< "$(curl -sSfk https://packages.local/service/rest/v1/repositories | jq -r '.[] | .["name"]' | grep ^csm-noos | tr '\n' ' ')"
+
+  echo "Will look for smart-mon rpm in the following repos: ${repos[@]}"
```
rustydb: Apologies, gave you the wrong reference. Should be `*` to reference `$repos` as a string.
```suggestion
  echo "Will look for smart-mon rpm in the following repos: ${repos[*]}"
```

rustydb: https://www.shellcheck.net/wiki/SC2145

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,113 @@
...
+      echo "ERROR on line $LINENO: unable to retreive continuation token, exiting"
+      return 1
+    fi
+  done
+}
+
+get_smartmon_url
+if [[ $? -ne 0 ]]; then
+  exit 1
+fi
```
rustydb: Check exit code directly, this will help when `set -e` is set. Otherwise if line 98 failed, the script would exit before running line 99.
```suggestion
if ! get_smartmon_url; then
  exit 1
fi
```

scripts/operations/ceph/enable-smart-mon-storage-nodes.sh
```diff
@@ -0,0 +1,113 @@
...
+
+  # search through the csm-noos repos looking for our package
+  for repo in "${repos[@]}"; do
+    # Retrieve the packages from nexus
+    if [[ -z "$smartmon_url" ]]; then
+      smartmon_url=$(paginate "https://packages.local/service/rest/v1/components?repository=$repo" \
+        | jq -r '.items[] | .assets[] | .downloadUrl' | grep smart-mon | sort -V | tail -1)
+      if [[ $? -ne 0 ]]; then
+        return 1
+      fi
```
rustydb: Same comment as https://github.com/Cray-HPE/docs-csm/pull/4299/files#r1337850032.
```suggestion
      if ! smartmon_url="$(paginate "https://packages.local/service/rest/v1/components?repository=$repo" \
        | jq -r '.items[] | .assets[] | .downloadUrl' | grep smart-mon | sort -V | tail -1 )"; then return 1; fi
        return 1
      fi
```

mharding-hpe:
This also needs a backport to 1.6, right?

leliasen-hpe:
> This also needs a backport to 1.6, right?

This shouldn't need a backport. This is being introduced in CSM 1.5. Which is why it is important to install it before upgrading from CSM 1.4 to 1.5. However, it will already be installed by the time we upgrade to CSM 1.6 so it shouldn't be needed in an upgrade from CSM 1.5 to 1.6.

rustydb (approved):

PR #4323, opened 2023-09-28
CASMINST-6679: Use explicit cray-sat container image version

haasken-hpe:
# Description

Use an explicit version tag for the cray-sat container image. The 3.25.2 version is the version included in the latest release/1.5 branch of the csm repository. We can run into problems with caching of the image on the Kubernetes nodes if we use the `csm-latest` tag.

An alternative to this might be to use an imagePullPolicy of Always, but I'm not sure what other impacts that might have. Additionally, using an explicit version is what was done for the iuf container in the iufBase template, so do the same here for consistency.

# Checklist

### DIFF
--- a/workflows/templates/base/sat-general-iuf.template.argo.yaml
+++ b/workflows/templates/base/sat-general-iuf.template.argo.yaml
@@ -62,7 +62,7 @@ spec:
             valueFrom:
               path: "{{inputs.parameters.script_stdout_file}}"
       script:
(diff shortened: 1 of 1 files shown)


haasken-hpe:
This happens to be the latest published stable version of the cray-sat container image, so it is currently also the version targeted for CSM 1.6, so I'm going to use the "/backport" action to port it to the release/1.6 branch as well.

bo-quan (approved):

bklei (approved):

heemstra (approved):

PR #4401, opened 2023-10-19
CASMTRIAGE-6182: Install latest RPMs from tarball onto m001 after redeploy

mharding-hpe:
# Description
Modify the install procedure so that the latest RPMs from the tarball are saved off and installed on m001 after redeploy.

For the same reasons as [CASMINST-6699](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6699), this only needs a backport for 1.4: https://github.com/Cray-HPE/docs-csm/pull/4402

# Checklist

### DIFF
--- a/install/scripts/install-goss-tests.sh
+++ b/install/scripts/install-goss-tests.sh
@@ -2,7 +2,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2021-2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -22,6 +22,9 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 #
+
+# This script is a replacement for the script of the same name in the lib
+# directory of the CSM tarballs and repository.
 
 # This script is run twice during CSM installs. It does different things depending
 # on when it is run.
--- a/install/scripts/install-goss-tests.sh
+++ b/install/scripts/install-goss-tests.sh
@@ -1,0 +1,201 @@
+#!/bin/bash
+#
+# MIT License
+#
+# (C) Copyright 2023 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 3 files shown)


install/scripts/install-goss-tests.sh
rustydb, line 1: Please enlighten me, what's the reason that this script is added to the repo instead of invoking the one from the tarball?

https://github.com/Cray-HPE/csm/blob/release/1.6/lib/install-goss-tests.sh

I can understand how deploy_final_non-compute_node doesn't have access to the tarball, but at this point in the install csm-testing exists in nexus and will be installed by cloud-init on m001 through the `install_csm_rpms` function during `runcmd`.

I feel like I am missing something, and just want to be sure we're fixing the right thing.

mharding-hpe: As you said, I cannot use the tarball script directly to handle `m001`, for the reasons you outlined.

In the case of the 1.3.5 bug, it is being caused by the `hpe-csm-goss-package` not being updated on `ncn-m001`. I didn't think `cloud-init` was covering that in 1.3 and 1.4. Am I mistaken about that?

For any RPMs which are covered by `cloud-init`, I can certainly modify this script to skip those on `m001`, if desired.

rustydb: Ok, that's where my uncertainty was coming from. I had lost context, and you're correct - `hpe-csm-goss-package` is not installed by the cloud-init function in 1.3 and 1.4 without an image change.

That being said, should we de-dup this script? If it lives here, can it be yanked out of the tarball? Otherwise, do we have to now update _both_ scripts going forward for 1.3 and 1.4?

mharding-hpe: The only reason I didn't do that is because I wasn't sure if any changes had been made to the csm tarball version of the script from 1.3.0, 1.3.1, etc, to 1.3.5. Obviously all 1.3 installs use the same docs RPM version, but they use that script version from the tarball. So basically I didn't want to risk accidentally breaking installs of older 1.3 versions.

Let me see what changes were made to that script between different csm 1.3.x and 1.4.x versions. If they don't seem like they'd be broken by doing it, I will modify my script to supersede them.

Oh, and also, it sounds like we don't want to respin the 1.3 build, in which case we wouldn't really be able to delete the script from the tarball. But I could modify the docs-csm script to replace the tarball script.

mharding-hpe: Ok, the changes between the 1.3.x versions seem minor, so I should be able to modify the docs-csm version of the script to supplant the tarball version. Stay tuned.

rustydb (approved):

nrockershousen (approved):

PR #4475, opened 2023-11-07
CASMPET-5760: Add network policy stashing, move restart-ds inline, and generate workflow

johren-hpe:
# Description

Last modifications for the cilium live migration workflow.   This adds generation of the live-migration workflow using a jinja2 template to generate for all of the masters and workers in the system.   At some point we may want to move this into cray-nls, but this is what I could get done for the current deadline.

Also adds stashing of the network policies.  This is required during the live migration process because we are changing NetworkPolicy providers from weave to cilium.   See https://docs.cilium.io/en/latest/installation/k8s-install-migration/

I also moved the restart daemonset process from a script to inline.  This avoids having to make changes for that process in the node image.

Relates to:
- [CASMPET-5760](https://jira-pro.it.hpe.com:8443/browse/CASMPET-5760)

# Checklist


workflows/templates/cilium.migrate.yaml
```diff
@@ -99,7 +99,13 @@ spec:
               - name: scriptContent
                 value: |
                   echo "restarting daemonsets on {{inputs.parameters.targetNcn}}"
-                  /srv/cray/scripts/common/restart-ds.sh {{inputs.parameters.targetNcn}}
+                  NODE={{inputs.parameters.targetNcn}}
+                  kubectl get nodes --no-headers=true | awk '{print $1}' | grep -q "^${NODE}$" || { echo "ERROR: Invalid node name ${NODE}"; exit 1; }
```
dborman-hpe: This line will work, but it could be simplified to:
```
kubectl get node ${NODE} >/dev/null 2>&1 || { echo "Error: Invalid node name ${NODE}; exit 1; }
```

workflows/templates/cilium.migrate.yaml
```diff
@@ -99,7 +99,13 @@ spec:
...
+
+                  while read -r a b; do
+                    echo "Restarting pod $b"
+                    kubectl -n $a delete pod $b
```
dborman-hpe: This would be a little clearer with descriptive variable names, like `ns` and `pod` instead of `a` and `b`

workflows/templates/cilium.post-migrate.yaml
```diff
@@ -35,7 +35,27 @@ spec:
...
+            template: shell-script
+          arguments:
+            parameters:
+              - name: dryRun
+                value: "{{inputs.parameters.dryRun}}"
+              - name: scriptContent
+                value: |
+                  cm_found=0
+                  kubectl -n argo get cm stash-netpol || cm_found=1
+                  if [ $cm_found -eq 1 ]; then
```
dborman-hpe: Is this logic correct?  Shouldn't it be `&&` instead of `||`?  The way it is currently written `cm_found` will be set when the `kubectl` command fails.  But it can be simplified to one line:
```
    if kubectl -n argo get cm stash-netpol; then
```
Which can be simplified even more to just:
```
if kubectl -n argo get cm stash-netpol -o jsonpath='{.data.netpol\.yaml}' > /tmp/netpol-out.yaml; then
    kubectl apply -f /tmp/netpol-out.yaml
    kubectl -n argo delete cm stash-netpol
fi
```
Also, do you want to leave around `/tmp/netpol-out.yaml`? or should there be another line to remove it?

workflows/templates/cilium.deploy.yaml
```diff
@@ -105,4 +105,32 @@ spec:
...
+            template: shell-script
+          arguments:
+            parameters:
+              - name: dryRun
+                value: "{{inputs.parameters.dryRun}}"
+              - name: scriptContent
+                value: |
+                  kubectl get netpol -A
+                  rm -f /tmp/netpol.yaml
+                  for i in $(kubectl get netpol -A --no-headers | awk '{print $1}' | sort | uniq);do
```
dborman-hpe: You can use `sort -u` instead of `sort | uniq`

workflows/templates/cilium.migrate.yaml
johren-hpe: Fixed

workflows/templates/cilium.post-migrate.yaml
johren-hpe: Fixed

workflows/templates/cilium.deploy.yaml
johren-hpe: Changed

workflows/templates/cilium.migrate.yaml
johren-hpe: Fixed

dborman-hpe (approved):
Thanks for addressing my comments, looks good to me.

rustydb (approved):

PR #4517, opened 2023-11-28
CASMTRIAGE-6263

Srinivas-Anand-HPE:
# Description
  Resolves CASMTRIAGE-6263 
  
  During an IUF management-nodes-rollout with csm 1.5, it was noted that tests that determine if the node can rebuilt are 
  not prefixed so they go to the user (INFO|WARN|ERROR). The CLI simply says the operation failed. IIRC there were no useful 
  msgs in the Argo UI either.

### DIFF
--- a/upgrade/scripts/upgrade/ncn-upgrade-master-nodes.sh
+++ b/upgrade/scripts/upgrade/ncn-upgrade-master-nodes.sh
@@ -62,7 +62,7 @@ if [[ $state_recorded == "0" ]]; then
   } >> ${LOG_FILE} 2>&1
   record_state "${state_name}" ${target_ncn}
 else
-  echo "====> ${state_name} has been completed"
+  echo "INFO ====> ${state_name} has been completed"
 fi
 
 if [[ ${target_ncn} == "ncn-m001" ]]; then
@@ -75,7 +75,7 @@ if [[ ${target_ncn} == "ncn-m001" ]]; then
     } >> ${LOG_FILE} 2>&1
     record_state "${state_name}" ${target_ncn}
   else
-    echo "====> ${state_name} has been completed"
+    echo "INFO ====> ${state_name} has been completed"
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


atifsyedali (approved):

upgrade/scripts/upgrade/ncn-upgrade-master-nodes.sh
```diff
@@ -205,14 +205,31 @@ if [[ $state_recorded == "0" ]]; then
...
+        echo "ERROR failed goss test: $line"
+      else
+        echo "$line"
+      fi
+    done
+  } | tee -a ${LOG_FILE} 2>&1
+  record_state "${state_name}" ${target_ncn}
+else
+  echo "INFO ====> ${state_name} has been completed"
+fi
```
leliasen-hpe: This should be replaced with the following.
```suggestion
state_name="RUN_GOSS_TESTS"
state_recorded=$(is_state_recorded "${state_name}" ${target_ncn})
if [[ $state_recorded == "0" ]]; then
  echo "====> ${state_name} ..."
  {
    set +e
    return_code=0
    while read line; do
      if [[ -n $(echo $line | grep 'Title\|desc') ]]; then
        echo "ERROR failed goss test: $line"
	return_code=1
      else
        echo -e "$line"
      fi
done < <(ssh $target_ncn -t "GOSS_BASE=/opt/cray/tests/install/ncn goss -g /opt/cray/tests/install/ncn/suites/ncn-upgrade-tests-master.yaml --vars=/opt/cray/tests/install/ncn/vars/variables-ncn.yaml validate")
    if [[ $return_code -eq 1 ]]; then
      exit $return_code
    fi
  } | tee -a ${LOG_FILE} 2>&1
  record_state "${state_name}" ${target_ncn}
else
  echo "INFO ====> ${state_name} has been completed"
fi
```
Changes are 
- set +e: so the segment doesn't exit if goss tests return an error
- read lines directly from goss test output. The goss_output variable was not receiving all of the output from the goss tests. 
- add a return_code so that the segment will exit if goss tests don't pass.

leliasen-hpe (changes requested):

rustydb (approved):

PR #4536, opened 2023-12-01
CASMTRIAGE-6273 : Adding a note to inform about the time consumption for session check

shivaprasad-metimath:
IM:CASMTRIAGE-6273
Reviewer:Ryan

# Description
It takes longer time to check the BOS sessions, when there are many of them. Hence a note is added to inform the user

# Checklist

### DIFF
--- a/operations/power_management/Prepare_the_System_for_Power_Off.md
+++ b/operations/power_management/Prepare_the_System_for_Power_Off.md
@@ -249,7 +249,7 @@ HPE Cray EX System Admin Toolkit (SAT) product stream documentation (`S-8031`) f
 
 1. (`ncn-mw#`) Check for running sessions.
 
-   > *Note:* This step may take a longer time if there are many BOS sessions. Recommend to keep the sessions count Minimal to save overall time taken.
+   > *Note:* This step may take a longer time if there are many BOS sessions running. It is recommended to keep the sessions count minimal to reduce the overall time taken.
 
     ```bash
     sat bootsys shutdown --stage session-checks 2>&1 | tee -a sat.session-checks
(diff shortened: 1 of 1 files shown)


rustydb (approved):

operations/power_management/Prepare_the_System_for_Power_Off.md
```diff
@@ -249,6 +249,8 @@ HPE Cray EX System Admin Toolkit (SAT) product stream documentation (`S-8031`) f
 
 1. (`ncn-mw#`) Check for running sessions.
 
+   > *Note:* This step may take a longer time if there are many BOS sessions. Recommend to keep the sessions count Minimal to save overall time taken.
```
nrockershousen:
```suggestion
   > *Note:* This step may take a longer time if there are many BOS sessions running. It is recommended to keep the sessions count minimal to reduce the overall time taken.
```

nrockershousen (approved):

operations/power_management/Prepare_the_System_for_Power_Off.md
```diff
@@ -249,6 +249,8 @@ HPE Cray EX System Admin Toolkit (SAT) product stream documentation (`S-8031`) f
 
 1. (`ncn-mw#`) Check for running sessions.
 
+   > *Note:* This step may take a longer time if there are many BOS sessions running. It is recommended to keep the sessions count minimal to reduce the overall time taken.
```
haasken-hpe, line 252: Hey, @rustydb and @nrockershousen, I was actually reviewing this and looking at the underlying cause of the slowness. A couple comments:

1. The edit to add the "running" qualifier on the end of "many BOS sessions" is incorrect. They do not have to be running. They just have to exist on the system because the code in `sat bootsys` is checking all existing sessions to see if they are running.
2. This note should not be needed in CSM 1.6. In that version, we are using BOS v2, which does not have the slowness issue because it checks session status in a more efficient way. I'd suggest that we only make this change in documentation for CSM v1.3 (the last version for which BOS v1 was the default). Adding it unnecessarily in CSM 1.6 just adds more confusion to an already confusing shutdown process.

Can we revert here and have @shivaprasad-metimath make a change like this only to CSM 1.3 docs?

PR #4618, opened 2024-01-16
CASMCMS-8841: Allow further granularity for controlling the boot progression for individual components.

jsl-hpe:
Document how BOS uses enable/disable for specifying which nodes to operate on and how to temporarily interrupt its' behavior.

# Description

# Checklist

### DIFF
--- a/operations/boot_orchestration/Components.md
+++ b/operations/boot_orchestration/Components.md
@@ -39,7 +39,7 @@ Stores information on the eventual desired boot artifacts and configuration for 
 
 ### `enabled`
 
-If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Disabled nodes may receive status updates from booted nodes, but BOS will not issue power commands to the nodes while they're disabled.
+If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Disabled nodes may receive status updates from booted nodes, but BOS will not issue power commands to the nodes while they are disabled.
 
 ### `error`
 
--- a/operations/boot_orchestration/Components.md
+++ b/operations/boot_orchestration/Components.md
@@ -39,7 +39,11 @@ Stores information on the eventual desired boot artifacts and configuration for 
 
 ### `enabled`
 
+<<<<<<< HEAD
 If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state.
+=======
+     If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Components can be disabled at any time; disabled components will not have power actions applied to them.
+>>>>>>> c2995d917 (Allow further granularity for controlling the boot progression for individual components.)
 
 ### `error`
 
--- a/operations/boot_orchestration/Components.md
+++ b/operations/boot_orchestration/Components.md
@@ -39,11 +39,7 @@ Stores information on the eventual desired boot artifacts and configuration for 
 
 ### `enabled`
 
-<<<<<<< HEAD
-If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state.
-=======
-     If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Components can be disabled at any time; disabled components will not have power actions applied to them.
->>>>>>> c2995d917 (Allow further granularity for controlling the boot progression for individual components.)
+If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Disabled nodes may receive status updates from booted nodes, but BOS will not issue power commands to the nodes while they're disabled.
 
 ### `error`
 
--- a/operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
+++ b/operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
@@ -10,9 +10,8 @@ BOS component records offer a history of operations that have been applied to a 
 cray bos components describe x3000c0s19b4n0
 ```
 
-Additionally, nodes that are unable to boot may also be disabled to prevent further BOS power actions from being applied
-to a node. This is helpful if BOS is currently retrying power operations on a specific node, and you need to introspect
-the node environment interactively.
+Additionally, nodes that are unable to boot may also be disabled. BOS will not apply power actions to disabled nodes,
+allowing an administrator to debug the node environment interactively.
 
 ```bash
 cray bos components update --enabled False x3000c0s19b4n0
--- a/operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
+++ b/operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
@@ -4,7 +4,7 @@ A number of tools can be used to analyze and debug issues encountered during the
 
 ## BOS
 
-Boot orchestration services offers records of actions that have been applied to a node through their component records.
+BOS component records offer a history of operations that have been applied to a node.
 A history of applied actions and future actions can be used to determine corrective actions and controlling actions that
 enables administrators to introspect issues on a per node level.
 
--- a/operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
+++ b/operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
@@ -17,7 +17,7 @@ to a node. This is helpful if BOS is currently retrying power operations on a sp
 the node environment interactively.
 
 ```bash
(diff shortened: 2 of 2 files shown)


operations/boot_orchestration/Components.md
```diff
@@ -39,7 +39,7 @@ Stores information on the eventual desired boot artifacts and configuration for
 
 ### `enabled`
 
-If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state.
+If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Disabled nodes may receive status updates from booted nodes, but BOS will not issue power commands to the nodes while they're disabled.
```
mharding-hpe:
```suggestion
If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Disabled nodes may receive status updates from booted nodes, but BOS will not issue power commands to the nodes while they are disabled.
```

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,32 @@
 
 A number of tools can be used to analyze and debug issues encountered during the compute node boot process. The underlying issue and symptoms dictate the type of tool required.
 
+## `BOS`
```
mharding-hpe:
```suggestion
## BOS
```

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,32 @@
...
+
+Boot orchestration services offers records of actions that have been applied to a node through their component records.
```
mharding-hpe:
```suggestion
BOS component records offer a history of operations that have been applied to a node.
```

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,32 @@
...
+A history of applied actions and future actions can be used to determine corrective actions and controlling actions that
+enables administrators to introspect issues on a per node level.
```
mharding-hpe:
```suggestion
```

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,32 @@
...
+
+```bash
+cray bos components describe x3000c0s19b4n0
+```
+
+Additionally, nodes that are unable to boot may also be disabled to prevent further BOS power actions from being applied
+to a node. This is helpful if BOS is currently retrying power operations on a specific node, and you need to introspect
+the node environment interactively.
```
mharding-hpe:
```suggestion
Additionally, nodes that are unable to boot may also be disabled. BOS will not apply power actions to disabled nodes,
allowing an administrator to debug the node environment interactively.
```

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,32 @@
...
+
+```bash
+cray bos components update --enabled False x3000c0s19b4n0
+```
+
+Once the problem has been isolated and fixed, you can return the node back over to BOS so that it may resume its boot
+session. Note: when re-enabling a node, BOS may first power cycle it to return it to working order depending on elapsed
+time and BOS options.
```
mharding-hpe:
```suggestion
Once the problem has been isolated and fixed, re-enable the node in BOS. Note: when re-enabling a node, BOS may first power cycle it to return it to working order, depending on elapsed time and BOS options.
```

mharding-hpe (approved):
Looks good other than some nits.

operations/boot_orchestration/Components.md
```diff
@@ -39,7 +39,7 @@ Stores information on the eventual desired boot artifacts and configuration for
 
 ### `enabled`
 
-If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state.
+If the node is enabled (enabled == True), BOS will take action to make the actual state match the desired state. Disabled nodes may receive status updates from booted nodes, but BOS will not issue power commands to the nodes while they are disabled.
```
don-bahls-hpe, line 42: This may be obvious but might be worth saying "... while they are disabled (i.e. enabled == False)."

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,27 @@
...
 
+## BOS
+
+BOS component records offer a history of operations that have been applied to a node.
+
+```bash
+cray bos components describe x3000c0s19b4n0
+```
+
+Additionally, nodes that are unable to boot may also be disabled. BOS will not apply power actions to disabled nodes,
```
don-bahls-hpe, line 13: Is there any other BOS operation that can clear the "enabled = False" state, or does the administrator have sole control over that?

jsl-hpe: At the end of the day, the component record is served up by an API, so any software or user could change this, however, in practice, the only things that toggle the enabled/disabled flag on the BOS level are:
- Nodes completing boot will toggle from enabled to disabled
- New BOS boot sessions that include the affected components will move from disabled to enabled
- Any other user who decides to enable or disable operation on a given BOS component.

operations/boot_orchestration/Components.md
jsl-hpe: I had a more verbose version before but changed it during peer review while the PR was in WIP.

don-bahls-hpe (approved):

rbak-hpe (approved):

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
jsollom-hpe: @don-bahls-hpe , note that in this case enabled = True means that BOS is going to operate on it. enabled = False means BOS is not going to operate on it. But this is a dynamic state. Nodes are enabled when a new BOS session is created. That is the first step.

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,27 @@
...
+allowing an administrator to debug the node environment interactively.
```
jsollom-hpe, line 14: Do we need to say anything about the node being disable, so long as it does not become a part of a new session?

I suppose an alternate way to disable nodes is to do so in HSM. That would also keep the node from being operated on. I think it's more dangerous because HSM disabling has to be manually undone whereas just making the node a part of a new BOS session re-enables it. From the debugger's point of view though, the reverse is true. HSM-disablement is safer because it cannot be accidentally undone.

jsl-hpe: I had attempted to view the problem as what we could accomplish in terms of cfs specifically, but this is probably the fine grained hammer that we need. Besides; not to hash up previous regretful skeletons, but HSM's disable flag in the PCS world is on the honor system 😆

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
```diff
@@ -2,6 +2,27 @@
...
+
+```bash
+cray bos components update --enabled False x3000c0s19b4n0
+```
+
+Once the problem has been isolated and fixed, re-enable the node in BOS. Note: when re-enabling a node, BOS may first power cycle it to return it to working order, depending on elapsed time and BOS options.
```
jsollom-hpe, line 20: This may make sessions look weird. For example, if I asked to shut down all of the nodes, but one didn't. I'd disable it, so I could go poke at it. Once the other nodes shut down, the session completes. Re-enabling this guy will cause the session to start running again, assuming it hasn't been cleaned up. We knew that would be a problem in the way we designed it. So these are probably idle musings.

operations/boot_orchestration/Tools_for_Resolving_Boot_Issues.md
jsollom-hpe: Doesn't BOS always honor it though, or did we change that when we converted to PCS?

I missed the connection between this documentation and it being specifically related to CFS. Could you explain?

rustydb (approved):

jsollom-hpe (approved):

PR #4283, opened 2023-09-14
CASMCMS-8720: Update docs for CFS v3 and paging

rbak-hpe:
# Description

Updates the CFS docs for the v3 api and paging.  Also re-organizes and cleans up the CFS docs in general.

Relates to:
- CASMCMS-8720

# Checklist

- [X] If I added any command snippets, the steps they belong to follow the prompt conventions (see [example][1]).
- [X] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].
- [X] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

 
[1]: https://github.com/Cray-HPE/docs-csm/blob/main/introduction/documentation_conventions.md#using-prompts
[2]: https://github.com/Cray-HPE/teams

PR #4620, opened 2024-01-16
CASMCMS-8720: Update argo workflows for CFS v3 and paging

mtupitsyn:
# Description

Follow up after #4283 - accommodate change in `/apis/cfs/v3/components` response format in argo workflows.

# Checklist

### DIFF
--- a/workflows/ncn/hooks/after-each/wait-for-cfs-after-rebuild.yaml
+++ b/workflows/ncn/hooks/after-each/wait-for-cfs-after-rebuild.yaml
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2022-2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022-2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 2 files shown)


rbak-hpe:
This does effectively address the upgrade issue between CFS v2 and v3, but it's strange to me that you're calling the "list components" endpoint repeatedly for a single component.  Why not just describe the component and check its status? This works, its just less efficient on the CFS side.

mtupitsyn:
> This does effectively address the upgrade issue between CFS v2 and v3, but it's strange to me that you're calling the "list components" endpoint repeatedly for a single component. Why not just describe the component and check its status? This works, its just less efficient on the CFS side.

In this PR, I only update `jq` expression to match new response format. I did not intend to rewrite logic of this wait condition.

mharding-hpe (approved):

mharding-hpe:
Theoretically there could be an issue here if the CFS query being made returns more results than will fit in a single response, but I am assuming from the context that the query is limited to a small-ish number of xnames.

mharding-hpe:
> This does effectively address the upgrade issue between CFS v2 and v3, but it's strange to me that you're calling the "list components" endpoint repeatedly for a single component. Why not just describe the component and check its status? This works, its just less efficient on the CFS side.

Looking at this code, I think it was based on the code from the venerable [`apply_csm_configuration`](https://github.com/Cray-HPE/docs-csm/blob/92c0a77a05aa7511b0ed887889cdad2d2f6a235a/scripts/operations/configuration/apply_csm_configuration.sh#L308) shell script. In that script, it uses similar logic, only it is doing so for all of the NCNs. I am guessing that whoever originally authored these workflows just mimicked that without realizing that it is not the best way to do it when dealing with a single specific node.

rustydb (approved):

PR #4621, opened 2024-01-18
CASMTRIAGE-6465 Fix assumptions about replica counts as well as how many containers a pod may have.

mitchty:
# Description

The existing checks assume too much about how a statefulset is setup and fails for anything that doesn't have 3 replicas. Additionally we really shouldn't be hard coding any of this. Another option might be to check the rollout status but this check queries the statefulset for how many replicas are actively receiving traffic instead of hoping the new settings are applied. Its also what I tested to fix this so I'd rather not hack in another test if this is 1.5 only anyway.

The new logic is simply foreach namespace query the pg statefulsets for their replicas and how many replicas are ready. Iff ready != replicas, fail fast on the first statefulset to not be valid.

# Checklist

### DIFF
--- a/upgrade/scripts/upgrade/util/fix-postgres.sh
+++ b/upgrade/scripts/upgrade/util/fix-postgres.sh
@@ -54,12 +54,25 @@ function postgres_crd_updated() {
   fi
 }
 
-function postgres_pods_running() {
-  if [ "$(kubectl get pod -l application=spilo -A --no-headers | awk '{ print $3 ":" $4 }' | sort -u)" == "3/3:Running" ]; then
-    return 0
-  else
-    return 1
-  fi
+# Query the statefulsets status, replicas = goal, readyReplicas = how many pods
+# are receiving traffic.
+#
+# Iff replicas != readyReplicas the cluster isn't ready.
+#
+# We do this so as not to hard code expectations on how many replicas might be
+# configured for the pg cluster.
+function postgres_statefulsets_ready() {
+  ok=0
+  # for <<< quoting this is unnecessary
+  #shellcheck disable=SC2046
+  while read -r ns ss ready replicas; do
+    if [ "${replicas}" -ne "${ready}" ]; then
+      ok=$((ok + 1))
+      printf "statefulset %s in namespace %s ready %d != replicas %d\n" "${ss}" "${ns}" "${ready}" "${replicas}" >&2
+    fi
+  done <<< $(kubectl get statefulset -l application=spilo -A -o jsonpath='{range .items[*]}{.metadata.namespace} {.metadata.name} {.status.readyReplicas} {.status.replicas}{"\n"}{end}' | sort)
+
+  return ${ok}
 }
 
 function postgres_clusters_running() {
@@ -109,7 +122,7 @@ kubectl get postgresql -A -o json \
   done
 echo "Waiting for 300 seconds for postgres statefulsets rolling restart to commence ..."
 sleep 300
-wait_for "all postgres pods running" postgres_pods_running
+wait_for "all postgres statefulsets in ready state" postgres_statefulsets_ready
 wait_for "all postgres clusters in state Running" postgres_clusters_running
 echo "State of postgres clusters after operator upgrade:"
 kubectl get postgresql -A


mtupitsyn (approved):

bo-quan (approved):

upgrade/scripts/upgrade/util/fix-postgres.sh
```diff
@@ -54,12 +54,26 @@ function postgres_crd_updated() {
...
+      # Note ${ns} and ${ss} are already in a "" block, this shellcheck warning is pointless here.
+      #shellcheck disable=SC2086
+      if [ "$(kubectl get statefulset --namespace ${ns} ${ss} -o jsonpath='{.status.replicas}{"\n"}{.status.readyReplicas}{"\n"}' | sort -u | wc -l)" -ne 1 ]; then
+        return 1
+      fi
+    done
+  done
+
+  return 0
 }
```
dborman-hpe: That's a lot of `kubectl` calls.  This can be simplified to a single `kubectl` call:
```
function postgres_statefulsets_ready() {
  kubectl get statefulset -A -l application=spilo  -o jsonpath='{range .items[*]}{.status.replicas} {.status.readyReplicas}{"\n"}{end}' | sort -u | \
    while read rep ready; do
      if [ "$rep" != "$ready" ]; then
        return 1
      fi
    done
  return 0
}
```
When I timed the two, yours took 14 seconds, this one took about 0.5 seconds

mitchty: That presumes all the ready and replicas are equal for all statefulsets which isn't true as we have 3 and 2 respectively. That is the underlying reason for individual calls.

dborman-hpe: No, it doesn't presume anything.  It gets a list of all the ready/replicas pairs in one call instead of doing a `kubectl` call for each pair.  It then checks each pair looking to see if any of them don't match (the `| sort -u` is a minor optimization that isn't really needed).  It's doing the exact same thing as your code.

dborman-hpe: For example, on mug right now:
```
ncn-m001:~ # kubectl get statefulset -A -l application=spilo  -o jsonpath='{range .items[*]}{.status.replicas} {.status.readyReplicas}{"\n"}{end}' | sort -u
2 1
3 2
3 3
```
The check will indicate failure on the first one, because `2 != 1`

dborman-hpe: My apologies, I didn't fully check that the return value was correct.  With the `|` it was running in a subshell, so the failure wasn't getting returned properly.  Here's the version that doesn't use a subshell (and adds `-r` to the `read` to keep `spellcheck` happy):
```
function postgres_statefulsets_ready() {
  while read -r rep ready; do
    if [ "$rep" != "$ready" ]; then
      return 1
    fi
  done < <(kubectl get statefulset -A -l application=spilo  -o jsonpath='{range .items[*]} {.status.replicas} {.status.readyReplicas}{"\n"}{end}')
  return 0
}
```

rustydb: Please resolve the conversation/thread if you're done.

dborman-hpe: I would like to see the code changed to a more efficient version, here's a timing comparison run on mug:
```
ncn-m001:/tmp # /tmp/dab.t2
Original version

real	0m5.186s
user	0m13.490s
sys	0m3.219s
Failure

New version

real	0m0.244s
user	0m0.000s
sys	0m0.001s
Failure
```

dborman-hpe: @mitchty, though I would prefer that this PR be changed to the more efficient code that only makes one `kubectl` call, this is not code that is run on a regular basis so I won't insist on this change.  You can choose to either make the change or add a comment that you aren't going to change it (since this is an infrequently used code path).  In either case, then mark this conversation as `resolved` and the merge will be able to proceed.

mitchty: Fair points, though this only runs once on an upgrade so I wasn't too concerned with runtime. My only apprehension is if something is broken or not in Ready we'd need to do the same to track down which statefulset is at issue.

Let me test this out after breaking a vshasta setup quick.

dborman-hpe: Yes, if you want to print out which ones aren't ready, this will do that (and go through the whole list, not just stop at the first that isn't ready).
```
function postgres_statefulsets_ready() {
  ret=0
  while read -r ns ss rep ready; do
    if [ "$rep" != "$ready" ]; then
      ret=1
      echo "Stateful set $ns/$ss not ready: replicas: $rep, ready: $ready"
    fi
  done < <(kubectl get statefulset -A -l application=spilo  -o jsonpath='{range .items[*]}{.metadata.namespace} {.metadata.name} {.status.replicas} {.status.readyReplicas}{"\n"}{end}')
  return $ret
}
```

mitchty: Should be updated with an autosquash! commit to basically do the suggestion. Tested it out with another broken (by me) virtual cluster.

dborman-hpe: Thanks!  That looks good to me.

rustydb:
considering this an approval: https://github.com/Cray-HPE/docs-csm/pull/4621#discussion_r1463927080

PR #4682, opened 2024-02-09
CASMINST-6788: Update backup PVC size to match nexus PVC size

ndavidson-hpe:
# Description

This updates the PVC size of the nexus-backup PVC created in the nexus-backup script to match the size of the nexus-data PVC on creation. This allows for systems that may have boosted the size of the Nexus PVC to run the nexus backup without errors.

# Checklist

### DIFF
--- a/scripts/nexus-export.sh
+++ b/scripts/nexus-export.sh
@@ -42,6 +42,7 @@ if ((usedNexus * 3 > availRGW)); then
 fi
 
 echo "Creating PVC for Nexus backup, if needed"
+backupSpaceNeeded=$(kubectl get pvc -n nexus nexus-data -o jsonpath='{.status.capacity.storage}')
 if [[ "Bound" != $(kubectl get pvc -n nexus nexus-bak -o jsonpath='{.status.phase}') ]]; then
   cat << EOF | kubectl -n nexus create -f -
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


ndavidson-hpe:
Waiting on a second look to backport.

scripts/nexus-export.sh
```diff
@@ -54,8 +55,31 @@ spec:
...
+spec:
+  storageClassName: k8s-block-replicated
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: $backupSpaceNeeded
+EOF
+  else
+    echo "Exiting script."
```
leliasen-hpe: Do you want to add a more detail here? Such as "A nexus backup cannot be taken because the the Backup PVC size does not match the nexus-data PVC. Exiting script." And should it be an exit 1 since the script did not complete what it was supposed to?

leliasen-hpe (changes requested):

bo-quan (approved):

leliasen-hpe (approved):

spillerc-hpe (approved):

scripts/nexus-export.sh
```diff
@@ -54,8 +55,31 @@ spec:
...
+  name: nexus-bak
+spec:
+  storageClassName: k8s-block-replicated
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: $backupSpaceNeeded
+EOF
+  else
```
mitchty, line 79: Where is the n case? this just checks that the user said yes, if they say no shouldn't we let them fail?

I'd see this as yes/no/abort and check for yes/no cases with the latter being the fall through.

ndavidson-hpe: In my eyes its a yes or abort case. With no its going to fail silently creating the tar. Saying No here means they dont want to make a new backup. Maybe a better phrasing for the question is needed.

mitchty: oki doki works for me was just wondering if we wanted to let them fail in the no case or not.

ndavidson-hpe:
/backport release/1.5 release/1.4 release/1.3 release/1.2 release/1.0

rustydb (approved):

PR #4755, opened 2024-02-27
CASMTRIAGE-6700-release-1.4 source myenv file at beginning of CSM 1.4 upgade

leliasen-hpe:
# Description
The CSM storage node upgrade uses two different files to keep track of the state of the upgrade, instead of the 1 file it should use. The node upgrade state file should be `/etc/cray/upgrade/csm/csm-1.4.0/ncn-s00x` but instead `/etc/cray/upgrade/csm/csm-/ncn-s00x` is used sometimes when the script doesn't have the CSM_RELEASE varaible. Sourcing the myenv file at the beginning of the upgrade will ensure the correct file is used for the whole upgrade.

# Checklist


leliasen-hpe:
This change is not necessary because storage nodes are not upgraded in CSM 1.4. I am closing this issue. This problem does not exist in CSM 1.5+ because storage nodes are upgraded via argo and this problem doesn't exist.

PR #4868, opened 2024-03-21
CASMTRIAGE-6796 Remove `ip=hsn*:auto6` Kernel arguments

rustydb:
# Description

Remove parameters for the HSN interfaces during dracut to avoid upping them and potentially toggling their CARRIER flag.

This addresses a condition where cloud-init selects an undesirable interface during its init-local stage, where it iterates through NICs until it finds one with a CARRIER flag. Since the HSN does not provide DHCP/DNS for the management network, and may not even be active during an NCN deployment, if cloud-init selects an HSN NIC it will fail DHCP.

Contextually speaking, the network configuration that is obtained during dracut is not fully carried over into the rootfs after the initramFS is deallocated. The IP and routing assigned to the NIC will persist, but the DNS configuration (which is stored in `/etc/resolv.conf` in both the initramFS and rootFS) is not preserved. This means that until something DHCPs again in the rootfs, that IP connectivity may work but DNS resolution will not.

This has been happening unbeknownst to us, at least without repercussions for a long time, it has only mattered for CSM 1.5 after the completion of MTL-2000. MTL-2000 added a dependency on DNS before our `runcmd` scripts are executed. If cloud-init init-local fails to setup DNS, then cloud-init modules that depend on DNS will fail.

By removing these `ip=hsn*` parameters, we cease attempting to UP the HSN NICs. Thus the HSN NICs will not show a CARRIER flag, and cloud-init will omit them when searching for a usable NIC.

This does not need to be backported to 1.6, but it could for consistency so I will make one.

# Checklist

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -2,7 +2,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2021-2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2021-2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -1091,7 +1091,6 @@ if [[ ${state_recorded} == "0" && $(hostname) == "${PRIMARY_NODE}" ]]; then
   echo "====> ${state_name} ..." | tee -a "${LOG_FILE}"
   {
 
-
     # As boot parameters are added or removed, update these arrays.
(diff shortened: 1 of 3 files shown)


upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -1104,6 +1105,44 @@ if [[ ${state_recorded} == "0" && $(hostname) == "${PRIMARY_NODE}" ]]; then
       csi handoff bss-update-param --set "${bootparameter}"
     done
 
+  # Get a list of NCNs.
```
rustydb: I need to double-check this portion works before we can merge.

rustydb: Done.

rustydb:
Will fix spellcheck and other gates before merging. Looking at this again after my 11 AM CST presentation.

cfawcett-hpe (approved):

PR #5049, opened 2024-04-29
CASMTRIAGE-6939 move 'update ceph node-exporter config for SNMP counters' to happen after storage node upgrade

leliasen-hpe:
# Description

Updating node-exporter on storage nodes needs to happen after the storage node upgrade. It needs to happen after because the command is not valid in earlier versions of Ceph. Previously, this step was in 'Validate CSM Health During Upgrade". However, that happens before management-nodes-rollout which causes an error. This PR changes the documentation so that the step happens at the correct time.

# Checklist


ndavidson-hpe (approved):

dborman-hpe (approved):
LGTM

rustydb (approved):

PR #5109, opened 2024-05-22
CASMPET-7069 Timing issue with Nexus upgrade in prerequisites.sh

mtupitsyn:
# Description
Previously, `prerequisites.sh` script was updating `cray-precache-images` configmap and waiting for 120s for nexus images to be pulled on all nodes, prior to nexus upgrade. This appeared to be unreliable (120s was not enough) and unnecessary for just a single pull operation. Updating `cray=precache-images` configmap does not make much sense, because it is getting overwritten with chart upgrade soon after. Replaced this logic with ssh to each worker node and invoking `crictl pull` directly.

# Checklist

# Testing
Tested on vShasta upgrade.

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -521,8 +521,9 @@ if [[ ${state_recorded} == "0" && $(hostname) == "${PRIMARY_NODE}" ]]; then
     # Skopeo image is stored as "skopeo:csm-${CSM_RELEASE}"
     podman load -i "${CSM_ARTI_DIR}/vendor/skopeo.tar"
     nexus_images=$(yq r -j "${CSM_MANIFESTS_DIR}/platform.yaml" 'spec.charts.(name==cray-precache-images).values.cacheImages' | jq -r '.[] | select( . | contains("nexus"))')
+    worker_nodes=$(grep -oP "(ncn-w\w+)" /etc/hosts | sort -u)
     while read -r nexus_image; do
-      echo "Uploading $nexus_image ..."
+      echo "Uploading $nexus_image into Nexus ..."
       podman run --rm -v "${CSM_ARTI_DIR}/docker":/images \
         "skopeo:csm-${CSM_RELEASE}" \
         --override-os=linux --override-arch=amd64 \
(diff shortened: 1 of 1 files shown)


rustydb (approved):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -521,8 +521,9 @@ if [[ ${state_recorded} == "0" && $(hostname) == "${PRIMARY_NODE}" ]]; then
     # Skopeo image is stored as "skopeo:csm-${CSM_RELEASE}"
     podman load -i "${CSM_ARTI_DIR}/vendor/skopeo.tar"
     nexus_images=$(yq r -j "${CSM_MANIFESTS_DIR}/platform.yaml" 'spec.charts.(name==cray-precache-images).values.cacheImages' | jq -r '.[] | select( . | contains("nexus"))')
+    worker_nodes=$(grep -oP "(ncn-w\w+)" /etc/hosts | sort -u)
```
rustydb: nit: match digits instead of any character.
```suggestion
    worker_nodes=$(grep -oP "(ncn-w\d+)" /etc/hosts | sort -u)
```

rustydb: It might be better to get this list of nodes from HMS instead, but this should work. The only case where it might not work, is if a system has added a node and `/etc/hosts` hasn't been updated yet (since we don't recreate it on reboot, and nothing forces updating it during hardware add).

This would get all of the XNAMES, which should be SSHable.
```bash
if IFS=$'\n' read -rd '' -a NCN_XNAMES; then
  :
fi <<< "$(cray hsm state components list --role Management --type Node --format json | jq -r '.Components | map(.ID) | join("\n")')"
```

rustydb (changes requested):

upgrade/scripts/upgrade/prerequisites.sh
mtupitsyn: This will return all nodes, not only workers, right? It's ok (although unnecessary) to have k8s masters on the list, but other nodes (e.g. storage) will return failure on `crictl` call.

mtupitsyn: Parsing `/etc/hosts` is used in a few previous steps in this file (`UPDATE_SSH_KEYS`, `REPAIR_AND_VERIFY_CHRONY_CONFIG`), so in case if new node was not added properly to `/etc/hosts`, there will be problems anyway.

rustydb (approved):

nrockershousen (approved):

PR #5182, opened 2024-06-25
CASMINST-6902: Improve/automate PIT data backup

mharding-hpe:
This replaces a few manual steps from the "Deploy Final NCN" procedure with a script. The script does the same thing, but does so more carefully. I also modified it to include a few additional files in the backup (which was my original motivation for doing any of this -- the DR work required some additional things to be included)

I tested this on mug and verified that it worked as expected.

Backports:
1.5: https://github.com/Cray-HPE/docs-csm/pull/5183
1.4: https://github.com/Cray-HPE/docs-csm/pull/5184

I am not backporting further just because I won't really be able to test it, and the original motivation (for the DR work) isn't applicable prior to CSM 1.4

### DIFF
--- a/install/scripts/backup-pit-data.sh
+++ b/install/scripts/backup-pit-data.sh
@@ -1,0 +1,139 @@
+#!/bin/bash
+#
+# MIT License
+#
+# (C) Copyright 2024 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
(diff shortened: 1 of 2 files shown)


install/scripts/backup-pit-data.sh
```diff
@@ -0,0 +1,119 @@
...
+
+echo "Creating PIT backup tarfile"
+
+pushd "${TEMPDIR}"
+run_cmd tar -czvf "${PITDATA}/prep/logs/pit-backup-$(date +%Y-%m-%d_%H-%M-%S).tgz" --remove-files *
+popd
+run_cmd rmdir -v "${TEMPDIR}"
+
+echo "Copying files to ncn-m002"
+ssh ncn-m002 \
```
rustydb: Since we're fixing this up, should we also set this to find the first-master-hostname instead of hardcoding `ncn-m002` in? This would help in case a system had some reason that m002 was unavailable or if they used a different first master.

mharding-hpe: Sure thing. I'll also need to modify the corresponding section that copies it back, but should be easy enough

mharding-hpe: Done!

rustydb (approved):

PR #5313, opened 2024-08-21
CASMCMS-8979 - add documentation for IMS remote build node status.

dlaine-hpe:
# Description

Add documentation to the new IMS remote build node status API. This allows users to query the current state of the remote build nodes to verify they are capable of running jobs. This will help the user debug what is going on when jobs fail to start on remote nodes. 

https://jira-pro.it.hpe.com:8443/browse/CASMCMS-8979

# Checklist

### DIFF
--- a/operations/image_management/Configure_a_Remote_Build_Node.md
+++ b/operations/image_management/Configure_a_Remote_Build_Node.md
@@ -688,3 +688,77 @@ Expected output will be something like:
     }
 ]
 ```
+
+### Checking the status of remote build nodes
+
+Before a job is set up to be launched on a remote build node, IMS checks that it is
+ready to run remote jobs. If a node is registered as a remote build node, but is not
+able to run remote jobs, that node will not have remote jobs scheduled on it. The
+status of the remote build nodes may be queried. The status object will have a field
+`ableToRunJobs` that contains a boolean indicating if the node is able to run jobs.
+If the node is not able to run jobs, the status object will document the issue with
+running jobs on that node.
+
+* NOTE: a value of '-1' for `numCurrentJobs` indicates that the actual number of
+   current jobs on the node was not able to be determined.
+
+(`ncn-mw#`) To check the status of a particular remote build node:
+
+```bash
+cray ims remote-build-nodes status describe "${IMS_REMOTE_NODE_XNAME}"
+```
+
+If the node is ready to run jobs, the output will look something like:
+
+```text
+{
+  "ableToRunJobs": true,
+  "nodeArch": "x86_64",
+  "numCurrentJobs": 0,
+  "podmanStatus": "Podman present at /usr/bin/podman",
+  "sshStatus": "SSH connection established.",
+  "xname": "x3000c0s3b0n0"
+}
+```
+
+If the node is not able to run jobs, the output will look something like:
+
+```text
+{
+  "ableToRunJobs": false,
+  "nodeArch": "Unknown",
+  "numCurrentJobs": -1,
+  "podmanStatus": "Unknown",
+  "sshStatus": "Unable to connect to node. Error: [Errno -2] Name does not resolve",
+  "xname": "x3000c0s3b0n1"
+}
+```
+
+(`ncn-m-w#`) To check the status of all defined remote build nodes:
+
+```bash
+cray ims remote-build-nodes status list
+```
+
+The output will look something like:
+
+```text
+[
+  {
+    "ableToRunJobs": true,
+    "nodeArch": "x86_64",
+    "numCurrentJobs": 2,
+    "podmanStatus": "Podman present at /usr/bin/podman",
+    "sshStatus": "SSH connection established.",
+    "xname": "x3000c0s3b0n0"
+  },
+  {
+    "ableToRunJobs": false,
+    "nodeArch": "Unknown",
+    "numCurrentJobs": -1,
+    "podmanStatus": "Unknown",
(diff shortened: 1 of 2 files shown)


jsl-hpe (approved):

operations/image_management/Configure_a_Remote_Build_Node.md
```diff
@@ -688,3 +688,77 @@ Expected output will be something like:
...
+
+Before a job is set up to be launched on a remote build node, IMS checks that it is
+ready to run remote jobs. If a node is registered as a remote build node, but is not
+able to run remote jobs, that node will not have remote jobs scheduled on it. The
+status of the remote build nodes may be queried. The status object will have a field
+`ableToRunJobs` that contains a boolean indicating if the node is able to run jobs.
+If the node is not able to run jobs, the status object will document the issue with
+running jobs on that node.
+
+* NOTE: a value of '10000' for `numCurrentJobs` indicates that the actual number of
```
jsollom-hpe: 10,000 seems weird and slightly unintuitive. I'd look at that and go, "Woah, busy node."

operations/image_management/Configure_a_Remote_Build_Node.md
```diff
@@ -688,3 +688,77 @@ Expected output will be something like:
...
+  "ableToRunJobs": true,
+  "nodeArch": "x86_64",
+  "numCurrentJobs": 0,
+  "podmanStatus": "Podman present at /usr/bin/podman",
+  "sshStatus": "SSH connection established.",
+  "xname": "x3000c0s3b0n0"
+}
+```
+
+If the node is not able to run jobs, the output will look something like:
```
jsollom-hpe: So, all of those are bad values, but any one of the could result in ableToRunJobs being set to 'false' , correct? 
If so, I think you could make that clearer.

jsollom-hpe (changes requested):

operations/image_management/Configure_a_Remote_Build_Node.md
dlaine-hpe: I am not sure what you mean by 'clearer'. The intent is to provide a reason for the user so they can go fix it. The 'unknown' values are because they can't be queried if we can't ssh into the node. This is a struct so I can't drop the values from the return object.

operations/image_management/Configure_a_Remote_Build_Node.md
dlaine-hpe: Yeah, I struggled a little on this one but the main point is to draw attention to the node.

dlaine-hpe: You are right - changed it to '-1' to signify unknown.

operations/image_management/Configure_a_Remote_Build_Node.md
jsollom-hpe: There's a certain amount of irony in my asking for clarity and not providing you with enough information to make the change.

What I meant by clearer is that I thought the documentation should say that _any one of these values_ in the structure below is enough to make the node ineligible to run jobs. When you say that it might look something like this, and then the node has four different things wrong with it, it could convey to the reader that an ineligible node must have all four of those things wrong with it. When, in fact, an ineligible node could actually look like this

{
  "ableToRunJobs": false
  "nodeArch": "unknown",
  "numCurrentJobs": 0,
  "podmanStatus": "Podman present at /usr/bin/podman",
  "sshStatus": "SSH connection established.",
  "xname": "x3000c0s3b0n0"
}

or this

{
  "ableToRunJobs": false,
  "nodeArch": "x86_64",
  "numCurrentJobs": -1,
  "podmanStatus": "Podman present at /usr/bin/podman",
  "sshStatus": "SSH connection established.",
  "xname": "x3000c0s3b0n0"
}
or this
{
  "ableToRunJobs": false,
  "nodeArch": "x86_64",
  "numCurrentJobs": 0,
  "podmanStatus": "Unknown",
  "sshStatus": "SSH connection established.",
  "xname": "x3000c0s3b0n0"
}

or this

{
  "ableToRunJobs": false,
  "nodeArch": "x86_64",
  "numCurrentJobs": 0,
  "podmanStatus": "Podman present at /usr/bin/podman",
  "sshStatus": "Unable to connect to node. Error: [Errno -2] Name does not resolve",
  "xname": "x3000c0s3b0n0"
}

Although I suspect with this last condition, that the podmanStatus would also be unknown as the node could not be reached. Possibly the numCurrentJobs and nodeArch might also be unknown, but I don't know if those are determined by introspecting the node, or if you keep that information separate in a database somewhere.

operations/image_management/Configure_a_Remote_Build_Node.md
```diff
@@ -688,3 +688,106 @@ Expected output will be something like:
...
+
+```text
+{
+  "ableToRunJobs": false,
+  "nodeArch": "x86_64",
+  "numCurrentJobs": -1,
+  "podmanStatus": "Podman not installed on node.",
+  "sshStatus": "SSH connection established.",
+  "xname": "x3000c0s3b0n1"
+}
```
jsollom-hpe, line 763: Do you need a fourth example where the number of jobs could not be determined?

jsollom-hpe (approved):

operations/image_management/Configure_a_Remote_Build_Node.md
dlaine-hpe: No - it will still run on a node where the number of jobs can't be determined.

PR #5457, opened 2024-10-11
CASMTRIAGE-7348 add script and documentation to add OSDs to removable disks

leliasen-hpe:
# Description

On Gamora, it was observed that each storage node only had 2 osds. Storage nodes are expected to have 4+ OSDs. On Gamora, the problem was that 4 disks on each storage node were "removable" and were not discoverable by the Ceph Orchestrator. In order to create OSDs on removable disks, I have added a script as well as troubleshooting documentation.

This script has been tested on Gamora.

# Checklist


jacobsalmela (approved):
I am not an expert in this area, but this looks reasonable.  I think this could also make a good test going forward.

nrockershousen (approved):

don-bahls-hpe:
One thing I noticed on the system where this script was run is that the REWEIGHT value for all of the new OSDs is set to 1.000 rather than .800 (or .799).   Do we need to adjust those values down to get ceph some space to shuffle data around?

dborman-hpe (approved):
LGTM

PR #5512, opened 2024-11-04
CASMINST-7038: After upgrading CSM services, upgrade the test/CLI RPMs on NCNs

mharding-hpe:
[CAST-36752](https://jira-pro.it.hpe.com:8443/browse/CAST-36752) was opened because during an upgrade from CSM 1.4 to 1.5.1, when they ran the CSM health validation after upgrading services, it failed because the version of the CLI on the NCNs had not yet been updated.

We have a step to do this during our patch upgrade process, but it appears to be overlooked curing the regular upgrade process.

This PR adds a call to our existing "RPM update" script at the tail end of the CSM service upgrade, to address this issue.

1.6 backport: https://github.com/Cray-HPE/docs-csm/pull/5513
1.4 backport (of just the upgrade-test-rpm.sh script changes): https://github.com/Cray-HPE/docs-csm/pull/5521

### DIFF
--- a/upgrade/scripts/upgrade/csm-upgrade.sh
+++ b/upgrade/scripts/upgrade/csm-upgrade.sh
@@ -156,8 +156,20 @@ fi
 # Restart CFS deployments to avoid CASMINST-6852
 "${basedir}/../common/restart-cfs.sh"
 
-# Update test/CLI RPMs on NCNs
-"${basedir}/util/upgrade-test-rpms.sh"
+state_name="UPDATE_TEST_CLI_RPMS"
+#shellcheck disable=SC2046
+state_recorded=$(is_state_recorded "${state_name}" $(hostname))
+if [[ $state_recorded == "0" ]]; then
+  echo "====> ${state_name} ..."
+  {
+    # Update test/CLI RPMs on NCNs
+    "${basedir}/util/upgrade-test-rpms.sh"
+  } >> ${LOG_FILE} 2>&1
+  #shellcheck disable=SC2046
+  record_state ${state_name} $(hostname)
+else
+  echo "====> ${state_name} has been completed"
+fi
 
 state_name="POST CSM Upgrade Validation"
 echo "====> ${state_name} ..."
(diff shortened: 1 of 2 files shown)


nrockershousen (approved):

upgrade/scripts/upgrade/csm-upgrade.sh
```diff
@@ -156,6 +156,12 @@ fi
 # Restart CFS deployments to avoid CASMINST-6852
 "${basedir}/../common/restart-cfs.sh"
 
+# Update test/CLI RPMs on NCNs
+"${basedir}/util/upgrade-test-rpms.sh"
+
+# Wait 10 seconds to make sure all of the restarted Goss servers are started up on the NCNs
+sleep 10
```
rustydb: This is pretty arbitrary, can we just check if the service has started instead or is that too difficult for this context?

mharding-hpe: Since it would be checking it on all of the remote servers, it seemed simpler to do it this way. But I guess I could just modify the upgrade-test-rpms script to do that.

jnowicki-hpe (approved):

jacobsalmela (approved):

upgrade/scripts/upgrade/csm-upgrade.sh
```diff
@@ -156,6 +156,9 @@ fi
 # Restart CFS deployments to avoid CASMINST-6852
 "${basedir}/../common/restart-cfs.sh"
 
+# Update test/CLI RPMs on NCNs
```
leliasen-hpe: Should a state be tracked for this? It won't hurt to do it multiple times but I imagine it would be best to only do it once. Specifically I mean doing the following

```bash
state_name="UPDATE_TEST_CLI_RPMS"
#shellcheck disable=SC2046
state_recorded=$(is_state_recorded "${state_name}" $(hostname))
if [[ $state_recorded == "0" ]]; then
  echo "====> ${state_name} ..."
  ...
```

mharding-hpe: Sure, I'm fine adding that. I forget why I chose not to do it in the first place.

mharding-hpe: Done!

leliasen-hpe (approved):

rustydb (approved):

PR #5537, opened 2024-11-14
CASMTRIAGE-7493: Provide troubleshooting/remediation for SBPS boot

aasha-hpe:
# Description

This is to provide troubleshooting/remediation for SBPS boot failure with images having no 'etag' or 'etag' set to NULL.
The issue reported in https://jira-pro.it.hpe.com:8443/browse/CASMTRIAGE-7489

# Checklist

- [NA] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].

 
### DIFF
--- a/troubleshooting/known_issues/SBPS_boot_fail.md
+++ b/troubleshooting/known_issues/SBPS_boot_fail.md
@@ -1,0 +1,156 @@
+# Boot content projection services can fail if an image does not have an `etag`
+
+## Issue Description
+
+iSCSI based boot content projection which is also known as "Scalable Boot Content Projection" (SBPS) for `rootfs` and `PE` images
+is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have an `etag` field or the `etag` field
+is present but set to NULL. Due to this problem with `etags`, the SBPS core service, the `SBPS Marshal agent`, crashes.
+
+## Issue Identification
+
+Issue can be identified with the symptoms below:
+
+Compute/UAN node (iSCSI Initiator) boot fails with the following messages in the console log:
+
+```text
+2024-11-12T20:07:50.17117 [  100.620313] dracut-pre-mount[3827]: iscsiadm: No active sessions.
+2024-11-12T20:13:00.44921 [  402.905538] dracut-pre-mount[5379]: ls: cannot access '/dev/mapper/594b7c79-9cb5-48f6-bceb-18f3596f0dbb_rootfs': No such file or directory
+2024-11-12T20:13:00.44925 [  402.924254] dracut-pre-mount[3814]: Warning: sbps-add-content.sh failed.
+2024-11-12T20:13:00.44926 [  402.952550] dracut-pre-mount[3809]: Warning: Unable to prepare squashfs file /tmp/cps/rootfs, dropping to debug.
+```
+
+SBPS Marshal agent crashes with the following messages on the worker node (iSCSI Target):
+
+```bash
+ncn-w002:~ # systemctl status sbps-marshal.service
+× sbps-marshal.service - System service that manages Squashfs images projected via iSCSI for IMS, PE, and other ancillary images similar to PE.
+     Loaded: loaded (/usr/lib/systemd/system/sbps-marshal.service; enabled; preset: disabled)
+     Active: failed (Result: exit-code) since Tue 2024-11-12 19:20:05 UTC; 18h ago
+   Duration: 2.232s
+    Process: 1239406 ExecStart=/usr/lib/sbps-marshal/bin/sbps-marshal (code=exited, status=1/FAILURE)
+   Main PID: 1239406 (code=exited, status=1/FAILURE)
+        CPU: 1.394s
+
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]:   File "/usr/lib/sbps-marshal/bin/sbps-marshal", line 8, in <module>
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]:     sys.exit(main())
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]:              ^^^^^^
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]:   File "/usr/lib/sbps-marshal/lib/python3.12/site-packages/bin/agent.py", line 270, in main
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]:     rootfs_s3_etag = artifact["link"]["etag"]
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]:                      ~~~~~~~~~~~~~~~~^^^^^^^^
+Nov 12 19:20:04 ncn-w002 sbps-marshal[1239406]: KeyError: 'etag'
+Nov 12 19:20:05 ncn-w002 systemd[1]: sbps-marshal.service: Main process exited, code=exited, status=1/FAILURE
+Nov 12 19:20:05 ncn-w002 systemd[1]: sbps-marshal.service: Failed with result 'exit-code'.
+Nov 12 19:20:05 ncn-w002 systemd[1]: sbps-marshal.service: Consumed 1.394s CPU time.
+```
+
+The complete log for the SBPS Marshal agent `systemd` service can be obtained from `journalctl` log as below:
+
+```bash
+    # journalctl -u sbps-marshal.service
+```
+
+## Workaround Description
+
+Step 1: Identify the manifest(s) which are missing the `etag` field
+
+This can be fetched from worker node as `s3:/boot-images` is mounted onto the worker node. The following command will
+search the mounted directory for manifests that are missing the `etag` field:
+
+```bash
+for f in /var/lib/cps-local/boot-images/*/manifest.json; do \
+    echo $f etag=$(cat $f | jq '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.etag' );  \
+done | grep etag=null
+```
+
+An example run of the command listing the problematic `manifest.json` is as follows:
+
+```bash
+# for f in /var/lib/cps-local/boot-images/*/manifest.json; do \
+     echo $f etag=$(cat $f | jq '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.etag' );  \
+ done | grep etag=null
+/var/lib/cps-local/boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/manifest.json etag=null
+```
+
+Looking at the `manifest.json`:
+
+```bash
+ncn-w006:~ # cat /var/lib/cps-local/boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/manifest.json
+{
+   "artifacts" : [
+      {
+         "link" : {
+            "path" : "s3://boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/compute.squashfs",
+            "type" : "s3"
+         },
+         "md5" : "fa69a6665124629bfb1d74bfa3e4b7de",
+         "type" : "application/vnd.cray.image.rootfs.squashfs"
+      }
+   ],
+   "created" : "20240815093713",
+   "version" : "1.0"
+}
+```
+
+The "link" section has no `etag` field.
+
+Step 2: Update the manifest with `etag` and update the same in `s3` from master node
+
+Example `manifest.json` update with `etag` obtained from `s3` and update the same in `s3`:
+
+```bash
+ncn-m001:~ # MANIFEST=$(cray ims images describe "3f89b76d-7578-4ac3-8891-0f8b4335b06d"  --format json | jq -r .link.path)
+ncn-m001:~ # echo $MANIFEST
+s3://boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/manifest.json
+
+ncn-m001:~ # cray artifacts get $(echo $MANIFEST | sed -e 's;^s3://;;g' -e 's;boot-images/;boot-images ;g' ) manifest.json
+
+ncn-m001:~ # cray artifacts describe $(cat manifest.json | jq -r '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.path' | sed -e 's;^s3://;;g' -e 's;boot-images/;boot-images ;g' ) --format json
+{
+  "artifact": {
+    "AcceptRanges": "bytes",
+    "LastModified": "2024-08-15T14:45:58+00:00",
+    "ContentLength": 22095331328,
+    "ETag": "\"f4153fe0a18cfb2205fb002d81a23ab2-2634\"",
+    "ContentType": "binary/octet-stream",
+    "Metadata": {
+      "md5sum": "fa69a6665124629bfb1d74bfa3e4b7de"
+    }
+  }
+}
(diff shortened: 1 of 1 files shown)


troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
...
+arch = "x86_64" 
+created = "2024-08-15T14:38:35.897457+00:00" 
+id = "3f89b76d-7578-4ac3-8891-0f8b4335b06d" 
+name = "kpreslan-20240815093713" 
+
+[link] etag = ""  <---------------------- `etag` set to NULL
+path = "s3://boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/manifest.json" 
+type = "s3" 
+
+[metadata]
```
dborman-hpe: This is not the problem, it's the missing "etag" in the manifest.json that causes the issue.

The first step is to provide the step to scan the system to identify whether or not there are any problematic manifest.json file.  Look in the description in the ticket, I added the information on how to identify which images have the issue.  Once you've identified the images, then describe how to fix them.

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
...
+Step 1: Check whether the image to be projected has `etag` in the image manifest
+
+example #1:
+
+```bash
+# cray ims images describe 3f89b76d-7578-4ac3-8891-0f8b4335b06d
+arch = "x86_64" 
+created = "2024-08-15T14:38:35.897457+00:00" 
+id = "3f89b76d-7578-4ac3-8891-0f8b4335b06d" 
+name = "kpreslan-20240815093713"
```
dborman-hpe: Don't include personally identifiable information.  In this output, replace `kpreslan` with something else, like `xxxxxxxx`.

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
+# Boot content projection to Compute/UANs via iSCSI fails if the image does not have a `etag`
+
+## Issue Description
+
+iSCSI based boot content projection which is also known as "Scalable Boot Content Projection" (SBPS) for `rootfs` and `PE` images
+is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have a `etag` field or `etag` field
```
jsollom-hpe:
```suggestion
is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have an `etag` field or the `etag` field

```

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
+# Boot content projection to Compute/UANs via iSCSI fails if the image does not have a `etag`
+
+## Issue Description
+
+iSCSI based boot content projection which is also known as "Scalable Boot Content Projection" (SBPS) for `rootfs` and `PE` images
+is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have a `etag` field or `etag` field
+is present but set to NULL. Due to this, the SBPS core service `SBPS Marshal agent` (runs as `systemd` service) crashes.
```
jsollom-hpe:
```suggestion
is present but set to NULL. Due to this problem with `etags`, the SBPS core service, the `SBPS Marshal agent`, crashes.

```

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
...
+
+iSCSI based boot content projection which is also known as "Scalable Boot Content Projection" (SBPS) for `rootfs` and `PE` images
+is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have a `etag` field or `etag` field
+is present but set to NULL. Due to this, the SBPS core service `SBPS Marshal agent` (runs as `systemd` service) crashes.
+
+## Issue Identification
+
+Issue can be identified with the symptoms below:
+
+Compute/UAN node (iSCSI Initiator) boot fails with the below messages in the console log:
```
jsollom-hpe:
```suggestion
Compute/UAN node (iSCSI Initiator) boot fails with the following messages in the console log:

```

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
...
+
+```text
+2024-11-12T20:07:50.17117 [  100.620313] dracut-pre-mount[3827]: iscsiadm: No active sessions.
+2024-11-12T20:13:00.44921 [  402.905538] dracut-pre-mount[5379]: ls: cannot access '/dev/mapper/594b7c79-9cb5-48f6-bceb-18f3596f0dbb_rootfs': No such file or directory
+2024-11-12T20:13:00.44925 [  402.924254] dracut-pre-mount[3814]: Warning: sbps-add-content.sh failed.
+2024-11-12T20:13:00.44926 [  402.952550] dracut-pre-mount[3809]: Warning: Unable to prepare squashfs file /tmp/cps/rootfs, dropping to debug.
+```
+
+SBPS Marshal agent crashes with the below messages on the worker node (iSCSI Target):
```
jsollom-hpe:
```suggestion
SBPS Marshal agent crashes with the following messages on the worker node (iSCSI Target):

```

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
...
+   "created" : "20240202221449",
+   "version" : "1.0"
+}
+```
+
+With the above examples given:
+ example #1 has the `etag` field set to NULL
+ example #2 has `etag` field missing.
+
+Step 2: Delete the image which has `etag` missing or set to NULL
```
jsollom-hpe: Is another possible solution to update the IMS record so that it does contain an etag field? Use the md5sum of the image and show that in as the etag. I believe that is what the etags are supposed to be for IMS. Even a fake etag may do the trick.

jsollom-hpe (changes requested):

troubleshooting/known_issues/SBPS_boot_fail.md
dborman-hpe: It's not the IMS record itself that is the problem, it's parsing the `manifest.json` that the IMS record points to that causes the problem when the `etag` doesn't exist in the `manifest.json` file.

jsollom-hpe: Thanks for the clarification. Regardless, I think the solution should be to fix the etag with a satisfactory value -- whatever that may be -- rather than deleting the images. People uploaded the images for a reason, and they're not going to want to delete them.

troubleshooting/known_issues/SBPS_boot_fail.md
aasha-hpe: @dborman-hpe Thanks for that clarification!.  I looked at the ticket, so do we want to suggest the step to add missing 'etag' to the manifest or delete the offending manifest ?

aasha-hpe: As you mentioned in the note, if we are suggesting to add missing 'etag' then we need to provide steps to add etag and what value of etag.

troubleshooting/known_issues/SBPS_boot_fail.md
don-bahls-hpe, line 6:
```suggestion
is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have an `etag` field. 

```

I don't really understand what "doesn't have an `etag` field or `etag` field" is supposed to mean.   If something doesn't have an etag field we know it doesn't have an etag field, right?   Or am I missing something?

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,114 @@
+# Boot content projection to Compute/UANs via iSCSI fails if the image does not have a `etag`
```
don-bahls-hpe:
```suggestion
# Boot content projection services can fail if an image does not have a `etag`
```

troubleshooting/known_issues/SBPS_boot_fail.md
don-bahls-hpe:
```suggestion
name = "my-debug-image-20240815093713" 

```

troubleshooting/known_issues/SBPS_boot_fail.md
don-bahls-hpe: Just to be clear the image being booted may not necessarily be the image that is causing the issue.   It could be any image.

troubleshooting/known_issues/SBPS_boot_fail.md
don-bahls-hpe: This is one way you can look for problematic images from a worker node

```
# from a worker node
for f in /var/lib/cps-local/boot-images/*/manifest.json; do 
    echo $f etag=$(cat $f | jq '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.etag' ); 
done | grep etag=null
```

troubleshooting/known_issues/SBPS_boot_fail.md
aasha-hpe: Ok I see Jason has added comment to add etag instead of deleting the images. But need exact steps/commands for the same.

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,150 @@
...
+         "type" : "application/vnd.cray.image.rootfs.squashfs"
+      }
+   ],
+   "created" : "20240815093713",
+   "version" : "1.0"
+}
+ncn-m001:~ # cray artifacts create $( echo $MANIFEST | sed -e 's/^s3:\/\///g' -e 's/boot-images\//boot-images /g' ) manifest.json
+```
+
+Step 3: Restart SBPS Marshal agent `systemd` service and ensure its running successfully
```
don-bahls-hpe:
```suggestion
Step 3: Restart SBPS Marshal agent `systemd` service on each impacted worker nodes and ensure its running successfully

```

don-bahls-hpe (approved):

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,150 @@
+# Boot content projection services can fail if an image does not have a `etag`
+
+## Issue Description
+
+iSCSI based boot content projection which is also known as "Scalable Boot Content Projection" (SBPS) for `rootfs` and `PE` images
+is supported in CSM 1.6.0. SBPS fails if the `rootfs` image manifest to be projected doesn't have an `etag` field or the `etag` field
+is present but set to NULL. Due to this problem with `etags`, the SBPS core service, the `SBPS Marshal agent` crashes.
```
jsollom-hpe:
```suggestion
is present but set to NULL. Due to this problem with `etags`, the SBPS core service, the `SBPS Marshal agent`, crashes.

```

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,150 @@
...
+    "ContentLength": 22095331328,
+    "ETag": "\"f4153fe0a18cfb2205fb002d81a23ab2-2634\"",
+    "ContentType": "binary/octet-stream",
+    "Metadata": {
+      "md5sum": "fa69a6665124629bfb1d74bfa3e4b7de"
+    }
+  }
+}
+ncn-m001:~ # cray artifacts describe $( cat manifest.json | jq -r '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.path' | sed -e 's/^s3:\/\///g' -e 's/boot-images\//boot-images /g' ) --format json | jq -r '.artifact.ETag' | tr -d '"'
+f4153fe0a18cfb2205fb002d81a23ab2-2634
```
dborman-hpe, line 121: Shouldn't there be a comment that at this point the `manifest.json` file is edited and the etag added?

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,150 @@
...
+
+```bash
+ncn-m001:~ # MANIFEST=$( cray ims images describe "3f89b76d-7578-4ac3-8891-0f8b4335b06d"  --format json | jq -r .link.path )
+ncn-m001:~ # echo $MANIFEST
+s3://boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/manifest.json
+
+ncn-m001:~ # cray artifacts get $( echo $MANIFEST | sed -e 's/^s3:\/\///g' -e 's/boot-images\//boot-images /g' ) manifest.json
+ncn-m001:~ # cat manifest.json
+{ "artifacts" : [ { "link" : { "path" : "s3://boot-images/3f89b76d-7578-4ac3-8891-0f8b4335b06d/compute.squashfs", "type" : "s3" }, "md5" : "fa69a6665124629bfb1d74bfa3e4b7de", "type" : "application/vnd.cray.image.rootfs.squashfs" } ], "created" : "20240815093713", "version" : "1.0" }
+ncn-m001:~ # cat manifest.json | jq -r '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.path' | sed -e 's/^s3:\/\///g' -e 's/boot-images\//boot-images /g'
```
jsollom-hpe: Why are we showing these two lines? I understand what they are doing. Does the reader need to see them? They're never actually using this command outright, nor are they using its output.

dborman-hpe:
```
sed -e 's/^s3:\/\///g' -e 's/boot-images\//boot-images /g'
```
This can be simplified to remove the backslashes by using a different character for the substitution separator:
```
sed -e 's;^s3://;;g' -e 's;boot-images/;boot-images ;g'
```
The same thing in several other places below.

troubleshooting/known_issues/SBPS_boot_fail.md
```diff
@@ -0,0 +1,150 @@
...
+    "ETag": "\"f4153fe0a18cfb2205fb002d81a23ab2-2634\"",
+    "ContentType": "binary/octet-stream",
+    "Metadata": {
+      "md5sum": "fa69a6665124629bfb1d74bfa3e4b7de"
+    }
+  }
+}
+ncn-m001:~ # cray artifacts describe $( cat manifest.json | jq -r '.artifacts[] | select(.type == "application/vnd.cray.image.rootfs.squashfs").link.path' | sed -e 's/^s3:\/\///g' -e 's/boot-images\//boot-images /g' ) --format json | jq -r '.artifact.ETag' | tr -d '"'
+f4153fe0a18cfb2205fb002d81a23ab2-2634
+ncn-m001:~ # cat manifest.json
```
jsollom-hpe, line 127: You never give an explicit instruction to update the manifest with the etag. You could say this.

Now update the manifest.json file with the newly discovered `etag` as show below.

troubleshooting/known_issues/SBPS_boot_fail.md
jsollom-hpe: nodes --> node
its --> it's

jsollom-hpe (changes requested):

troubleshooting/known_issues/SBPS_boot_fail.md
dborman-hpe: And `worker node` (singular)

troubleshooting/known_issues/SBPS_boot_fail.md
don-bahls-hpe: I was thinking when I came up with the example that you'd copy this to the next command, but then combined this into the next command, so it's not needed now as you noted.

jsollom-hpe (approved):

dborman-hpe (approved):
LGTM, thanks for the updates

PR #5651, opened 2025-01-13
CASMTRIAGE-7676 update the CSM 1.5.2 patch procedure to update the myenv file

leliasen-hpe:
# Description

A step needs to be added in the patch process to check that the myenv file is correct.

There are state files that track master node upgrades and steps do not run if they have already run successfully. The state files during this patch upgrade have not been working correctly. Only some steps have been written to the state file while other steps are written to an incorrect state file. A nested script gets the location of the state file that should be used from '/etc/cray/upgrade/csm/myenv'. The 'myenv' file was incorrect and thus it was making the master node upgrade not work correctly.

This PR updates the `myenv` file before the node upgrades in the patch procedure and thus the correct state file will be used and the upgrade should work properly.

# Checklist

### DIFF
--- a/upgrade/1.5.2/Upgrade_NCN_images.md
+++ b/upgrade/1.5.2/Upgrade_NCN_images.md
@@ -62,64 +62,50 @@ upgrade procedure which could cause problems.
     echo "${CSM_ARTI_DIR}"
     ```
 
-1. (`ncn-m001#`) Ensure the `/etc/cray/upgrade/csm/myenv` file has the correct values for `CSM_RELEASE` and `CSM_ARTI_DIR`.
+1. (`ncn-m001#`) Update the `/etc/cray/upgrade/csm/myenv` file with the correct `CSM_RELEASE`, `CSM_ARTI_DIR` and `CSM_REL_NAME` values.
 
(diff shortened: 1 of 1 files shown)


studenym-hpe (approved):

studenym-hpe:
I've got a general noobie question ... 
Why both ${CSM_RELEASE} (exported in the myenv file) and ${CSM_RELEASE_VERSION} (set to ${CSM_RELEASE} after sourcing the myenv file)?

leliasen-hpe:
> I've got a general noobie question ... Why both ${CSM_RELEASE} (exported in the myenv file) and ${CSM_RELEASE_VERSION} (set to ${CSM_RELEASE} after sourcing the myenv file)?

Almost everywhere in the CSM upgrade we use the CSM_RELEASE variable. For some reason, during the patch upgrade, we are exporting the CSM_RELEASE_VERSION variable. I don't believe we use the  CSM_RELEASE_VERSION variable during the master node upgrades, but I have exported it out of an abundance of caution. I don't want to impact the 1.5.2 patch procedure and cause it possibly to fail because I removed the variable. The CSM_RELEASE_VERSION is not a common variable in CSM and I don't want to add it to the myenv file.

bmcdonald3 (approved):

dborman-hpe (approved):
LGTM

PR #5703, opened 2025-01-25
CASMINST-7139: Update SAT bootprep known issue for 1.5.3

mharding-hpe:
There is a known SAT issue in CSM 1.5.2 that is also possible to hit in CSM 1.5.3. This PR updates the relevant workaround procedure and references to the procedure to account for the fact that it is present in both patch releases. The workaround remains essentially the same, and we tested the updated version on hela.

No backports needed -- this is unique to CSM 1.5.

The diff for the workaround procedure looks so huge because the version checked into the repository had Windows newlines, and they got stripped by git when I checked it in.


annapoorna-s-alt (approved):

ethanholen-hpe (approved):
Looks good to me 👍🏻

PR #5751, opened 2025-02-13
CASMPET-7385 add cleanup-live-images step before upgrading storage nodes

leliasen-hpe:
# Description

This PR adds a step to run the `cleanup-live-images.sh` right before upgrading storage nodes. 

This resolves the following two problems

1. If storage nodes have already been upgraded, if their node image is rebuilt again, when the "upgrade" to the new image is reattempted, it fails because the new SquashFS is not pulled since it already exists on the system. It was previously expected that rd.live.overlay.reset=1 would fix this however that does not work the same in SP6 images. 

2. On storage nodes, old squashfs images still exist after a node upgrade and take up a lot of space. This PR will remove these.

This PR also deletes the old storage node upgrade script which is no longer used.

**Testing**

This change has not been tested on a CSM 1.5 to CSM 1.6.

I have tested this on a CSM 1.6 storage node that is "upgraded" to different node image that is also CSM 1.6.

Successful Test:

image

Cleanup-live-images ran successfully right before node was rebooted. The boot and cloud-init ran successfully when it previously would have failed. Output from the cleanup-live-images stage is below. I have run this 3 times.

```text
Image storage status:
 
        Size: 23G
        Used: 4.6G
 
Current used image directory is: [/run/initramfs/live/1.6.1-rc.3]
Found the following unused image directories:
        1.5.0   2.2G
        1.6.1-rc.3      2.4G
-y was present; automatically removing images ...
This node will be unable to diskboot until it is reimaged with a netboot.
Removing squashFS directories ...
-a was present; removing ALL images including the currently booted image [/run/initramfs/live/1.6.1-rc.3]
Removing overlayFS directories ...
Done
Attempted to remount /run/initramfs/live as read-only but the device was busy. This will correct itself on the next reboot.
```

I reproduced the failure. Without this new cleanup step, the failure is seen every time a storage node is upgraded to an image with the same CSM version as the image currently running.

# Checklist


bmcdonald3 (approved):
Looks good to me

rustydb (approved):

PR #5986, opened 2025-05-27
CASMCMS-9442: Script linting

mharding-hpe:
Very minor linting of scripts. This cleans up issues I found when working on other things. The types of corrections are:

1. Remove trailing whitespace from lines
2. Remove unused function from `write_root_secrets_to_vault.py` (and removed related unused `traceback` import from same)
3. Correctly mark the `log_error_raise_exception` as `NoReturn` (without this, `pylint` will complain in some cases -- it has no impact at runtime)
4. Use string interpolation instead of f-strings when logging (`pylint` complains otherwise, because doing this allows you to avoid needless string formatting in cases where a message is not logged due to the log level)

None of these things change how any of these scripts work from an end-user perspective. But I wanted to make the changes separately from my main script PR work, to avoid cluttering it.

The PR does include one minor bug fix in the `write_root_secrets_to_vault.py` script. Specifically, it crashes if the user tries to run it when their public or private SSH key files do not exist, even if they choose options which should not require them to exist. This is because of a subtlety in how Python `argparse` handles defaults for arguments. This PR modifies how that script parses its arguments, to avoid this issue.

Backports:
1.6: https://github.com/Cray-HPE/docs-csm/pull/5987
1.5: https://github.com/Cray-HPE/docs-csm/pull/5988
1.4: https://github.com/Cray-HPE/docs-csm/pull/5989
1.3: https://github.com/Cray-HPE/docs-csm/pull/5990


dlaine-hpe (approved):

scripts/operations/node_management/delete_bmc_subscriptions.py
```diff
@@ -124,6 +124,6 @@ def clear_bmc_subscriptions(username, password, target_bmc):
     if creds is None:
         print("Failed to retrieve credentials from SCSD")
         sys.exit(1)
-    
+
     # Clear the subscriptions from the BMC
     clear_bmc_subscriptions(creds["Username"], creds["Password"], args.target_bmc)
```
dborman-hpe: It would be nice if the files that are missing the final `\n` could be fixed along with this whitespace cleanup.  Could you add that?  The list of files missing the final `\n` is:
```
[HPE-CG403QV9LF] for f in $(find . -type f -print); do if ! tail -1c $f | od -c | grep -q '\\n'; then echo $f; fi; done
./remove_orphaned_artifacts.py
./operations/configuration/python_lib/__init__.py
./operations/firmware/FASBackupImages.py
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/sls_reservations_schema.json
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/sls_subnets_schema.json
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/hold/sls_hardware.schema.json.hold
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/hold/sls_networks_schema.json.hold
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/hold/sls_subnet_schema.json.hold
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/hold/sls_full.schema.json.hold
./operations/node_management/Add_Remove_Replace_NCNs/sls_utils/schemas/sls_networks_schema.json
./operations/node_management/Add_Remove_Replace_NCNs/mac_collection_script.ipxe
./operations/node_management/delete_bmc_subscriptions.py
./csm_rbd_tool.tar.gz
[HPE-CG403QV9LF] 
```
(obviously the `*.gz` file doesn't need to be updated...)

mharding-hpe: Done (except for one of them, which was a size-0 file, and thus seems okay to have no newline)

dborman-hpe: Thanks!

dborman-hpe (approved):
LGTM

rustydb (approved):

PR #6083, opened 2025-06-26
CASMTRIAGE-8332 Fix for use case when rolebingding is not present

mtupitsyn:
# Description
Workaround for CASMTRIAGE-8332 tries to set rolebinding labels and annotations unconditionally. This fails if rolebinding is not present. This change add condition, labels and annotations are only set when rolebinding is present.

# Checklist


bmcdonald3 (approved):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -754,9 +754,11 @@ sleep 10
 
 # Workaround for CASMTRIAGE-8332 before upgrading postgres-operator for CSM 1.7.x
 {
-  kubectl label --overwrite rolebinding --namespace default postgres-pod app.kubernetes.io/managed-by=Helm || true
-  kubectl annotate --overwrite rolebinding --namespace default postgres-pod meta.helm.sh/release-name=cray-postgres-operator || true
-  kubectl annotate --overwrite rolebinding --namespace default postgres-pod meta.helm.sh/release-namespace=services || true
+  if kubectl get rolebinding --namespace default postgres-pod &> /dev/null; then
```
dborman-hpe, line 757: I had added the `|| true` to these statements so that they wouldn't fail the script if they failed, but you'd still see the error messages.  I'm okay with adding the check and just not doing the commands as a cosmetic fix so you don't see the error messages.
(This is just cosmetic, right?  Just to get rid of the error messages?)

dborman-hpe (approved):
LGTM

upgrade/scripts/upgrade/prerequisites.sh
mtupitsyn: The reason why both mitigations is on code is that we did these mitigations in parallel, and I merged your PR earlier and rebased this one. In my implementation I chose  conditional over `|| true` because it is indeed cleaner (does nto produce error messages which may be confusing for engineers). I think it's safe to have both mitigations in place.

studenym-hpe (approved):

PR #6094, opened 2025-06-30
CASMPET-7609: upgrade_k8s.sh log prefix

davidfluck-hpe:
# Description

Add an `INFO` prefix to all log lines in upgrade_k8s.sh to ensure that IUF displays log output from this script. We hardcoded this particular log level in the interest of time.

[CASMPET-7609](https://jira-pro.it.hpe.com:8443/browse/CASMPET-7609)

This was tested on `molly` by re-running the `deploy-product` IUF stage, which calls `upgrade_k8s.sh`. The script-specific log output is successfully captured by IUF, as expected.

### DIFF
--- a/upgrade/scripts/k8s/upgrade_k8s.sh
+++ b/upgrade/scripts/k8s/upgrade_k8s.sh
@@ -102,7 +102,8 @@ pause["1.31"]="3.10"
 pause["1.32"]="3.10"
 
 function prefix() {
-  echo -n "$(date --iso-8601=seconds) $(basename "$0")"
+  # We hardcode a log level of INFO for now to ensure IUF displays our logs.
+  echo -n "INFO $(date --iso-8601=seconds) $(basename "$0")"
 }
 
 function fix_sysconfig() {


davidfluck-hpe:
Converting this to a draft while I check something.

studenym-hpe:
Please add `--quiet` to the zypper calls.

studenym-hpe:
We want all kube and zypper output to be redirected to a log file ... will that happen in this PR or a different one?

ndavidson-hpe (approved):

davidfluck-hpe:
We've since decided not to redirect output to a file, since that was a proposed fix for an IUF issue that we now know is unrelated.

davidfluck-hpe:
> Please add `--quiet` to the zypper calls.

Does this still need to be done since we aren't worried about log volume? I would prefer to have more verbose output, in case we encounter any zypper issues along the way.

studenym-hpe:
> > Please add `--quiet` to the zypper calls.
> 
> Does this still need to be done since we aren't worried about log volume? I would prefer to have more verbose output, in case we encounter any zypper issues along the way.

I don't think we have to do this any longer since voluminous logging isn't causing the IUF/argo problem.

studenym-hpe (approved):

PR #6100, opened 2025-07-02
CASMTRIAGE-8378: delete upgrade file in post-upgrade cleanup

davidfluck-hpe:
# Description

Delete /etc/cray/kubernetes/upgrade file at the end of the post-upgrade cleanup process. BSS will ensure that this file doesn't get recreated on boot, but the file will continue to exist until a reboot, unless we delete it. This ensures that the file doesn't exist longer than it has to.

### DIFF
--- a/upgrade/scripts/upgrade/cleanup.sh
+++ b/upgrade/scripts/upgrade/cleanup.sh
@@ -31,3 +31,4 @@ TOKEN=$(curl -s -S -d grant_type=client_credentials \
 export TOKEN
 
 python3 /usr/share/doc/csm/upgrade/scripts/upgrade/cleanup.py
+rm -f /etc/cray/kubernetes/upgrade
--- a/upgrade/scripts/upgrade/cleanup.sh
+++ b/upgrade/scripts/upgrade/cleanup.sh
@@ -31,4 +31,7 @@ TOKEN=$(curl -s -S -d grant_type=client_credentials \
 export TOKEN
 
 python3 /usr/share/doc/csm/upgrade/scripts/upgrade/cleanup.py
-rm -f /etc/cray/kubernetes/upgrade
+
+# Remove the upgrade file from all NCN hosts.
+ncns=$(grep -oP 'ncn-\aw\d+' /etc/hosts | sort -u | tr -t '\n' ',')
+pdsh -S -b -w ${ncns} "rm -f /etc/cray/kubernetes/upgrade"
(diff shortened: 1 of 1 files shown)


upgrade/scripts/upgrade/cleanup.sh
```diff
@@ -31,3 +31,4 @@ TOKEN=$(curl -s -S -d grant_type=client_credentials \
 export TOKEN
 
 python3 /usr/share/doc/csm/upgrade/scripts/upgrade/cleanup.py
+rm -f /etc/cray/kubernetes/upgrade
```
studenym-hpe: This file exists on all master nodes ....
```
ncn-m001:/etc/cray/kubernetes # ll upgrade
-rw-r--r-- 1 root root 0 Jul  2 13:59 upgrade

ncn-m002:/etc/cray/kubernetes # ll upgrade
-rw-r--r-- 1 root root 0 Jul  1 20:54 upgrade

ncn-m003:~ # ll /etc/cray/kubernetes/upgrade
-rw-r--r-- 1 root root 0 Jul  1 22:03 /etc/cray/kubernetes/upgrade
```

so we should probably remove from all of the master nodes.

davidfluck-hpe: Good catch, thank you!

upgrade/scripts/upgrade/cleanup.sh
```diff
@@ -31,3 +31,7 @@ TOKEN=$(curl -s -S -d grant_type=client_credentials \
 export TOKEN
 
 python3 /usr/share/doc/csm/upgrade/scripts/upgrade/cleanup.py
+
+# Remove the upgrade file from all NCN hosts.
+ncns=$(grep -oP 'ncn-\aw\d+' /etc/hosts | sort -u | tr -t '\n' ',')
```
studenym-hpe: I think an "a" got into the regex - should be `ncn-\w\d+` otherwise `ncns` is empty.  Without the "a", the list of ncns will include storage nodes, which do not have the upgrade file.  You only want masters and workers here.

How about this instead?  
```
ncns=$(grep -oP 'ncn-[m|w]\d+' /etc/hosts | sort -u | tr -t '\n' ',')
```

Seems to work
```
ncn-m001: # ncns=$(grep -oP 'ncn-[m|w]\d+' /etc/hosts | sort -u | tr -t '\n' ',')
ncn-m001: # echo $ncns
ncn-m001,ncn-m002,ncn-m003,ncn-w001,ncn-w002,ncn-w003,ncn-w004,
ncn-m001: #
```

rustydb (approved):

mtupitsyn (approved):

PR #6112, opened 2025-07-09
CASMTRIAGE-8426: fix PREPARE_KUBEADM in prerequisites.sh

davidfluck-hpe:
# Description
Update prerequisites.sh to only run `PREPARE_KUBEADM` if we're on Kubernetes 1.24 (and if the step hasn't been run before), which implies an upgrade from 1.6 to 1.7. Otherwise, the step will fail when upgrading between patch versions on 1.7, because of kubeadm config file structure changes across Kubernetes versions.

`KUBERNETES_MINOR_VERSION` is read from the existing `version` file, which ought to be consistent across versions as well, because prerequisites are run before node rollout.

### DIFF
--- a/upgrade/scripts/upgrade/prerequisites.sh
+++ b/upgrade/scripts/upgrade/prerequisites.sh
@@ -36,6 +36,7 @@ trap 'err_report' ERR INT TERM HUP EXIT
 declare -a UNMOUNTS=()
 
 PRIMARY_NODE="ncn-m001"
+KUBERNETES_MINOR_VERSION=$(cut -d'.' -f2 /etc/cray/kubernetes/version)
 
 while [[ $# -gt 0 ]]; do
   key=$1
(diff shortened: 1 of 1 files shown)


ndavidson-hpe (approved):

upgrade/scripts/upgrade/prerequisites.sh
```diff
@@ -1294,7 +1298,7 @@ fi
 # Removing PodSecurityPolicy causes cascading restarts of many pods.
 state_name="REMOVE_PSP"
 state_recorded=$(is_state_recorded "${state_name}" "$(hostname)")
-if [[ ${state_recorded} == "0" ]]; then
+if [[ ${state_recorded} == "0" && ${KUBERNETES_MINOR_VERSION} -eq 24 ]]; then
```
studenym-hpe, line 1301: Good catch!  `kubectl delete psp --all=true` will error on k8s 1.32.

studenym-hpe (approved):

dborman-hpe (approved):
LGTM

PR #6126, opened 2025-07-16
CASMPET-7628: Create script to clear iSCSI playbook layer from NCN worker CFS component state

mharding-hpe:
CASMPET-7616 allows administrators to limit which worker NCNs are enabled for iSCSI. This is done using an HSM group. If the group is changed, then in order for the changes to take effect, the iSCSI CFS playbook needs to run on the workers. This will not happen automatically, because CFS will not realize that the HSM group has changed.

CASMPET-7619 is going to document the procedures to alter which worker NCNs are enabled for iSCSI. This means it will need to tell admins to re-run the iSCSI CFS layer on the worker nodes, after they've made their changes. However, there are some subtleties to consider.

The default way that we force CFS to re-run on a node is to clear its state entirely. This causes CFS to run through all of the layers in its desired configuration. This can take a long time, even if most of them have already run previously.

Another alternative is to update the states of the worker NCN components in CFS, and instead of clearing the states, just remove the iSCSI layer. This will cause CFS to create a session to only run that specific layer, which is what we want. The issue here is that this is pretty tedious, especially if done from the command line.

This PR addresses this concern by writing a script to do this for them. It leverages a lot of existing Python modules we've written for other purposes, so the API side of interacting with CFS (and HSM) didn't require many code changes. It mostly had to add in some logic to glue them together (as well as some minor linting and additions).

This script takes no arguments and does the following:
1. Gets a list of worker NCNs from HSM
2. Queries CFS to get the corresponding components and their current states
3. Does some sanity checking to make sure that what we got back from CFS looks right in general (no missing components, no duplicate components, etc).
4. Checks to make sure no components have pending status. If any do, exit the script. This indicates that CFS is in the process of configuring them, or will be soon. In this case, the user should wait until that is done before running the script.
5. Checks to see if any of the components are missing iSCSI layers in their state. If any are, a warning will be printed. This would only be the case if the node has never had CFS run before, or if the node has a desired configuration which does not include the iSCSI layer. These nodes will not be updated in CFS.
6. Updates all of the worker NCN components in CFS. All of them have their error counts set to 0, and have the iSCSI layers in their states removed (but the rest of the layers are not modified).

I have tested this on fanta and drax and verified that it works as expected (both for nodes that have the iSCSI layer and nodes that do not). I also verified that it correctly detected pending components and aborted.

There will be partial backports of this to earlier releases, but only for applicable linting or CFS/HSM module improvements. The main script will not be backported.

### DIFF
--- a/scripts/operations/configuration/refresh_worker_iscsi_config.py
+++ b/scripts/operations/configuration/refresh_worker_iscsi_config.py
@@ -112,23 +112,22 @@ def main() -> None:
 
     if no_iscsi_status:
         print("WARNING: The following workers have no iSCSI layer in their CFS "
-              f"component states; Only their error counts will be reset: {no_iscsi_status}")
+              f"component states; they will not be modified: {no_iscsi_status}")
     print("The following workers will have their CFS states updated and error "
           f"counts reset: {sorted(set(worker_comp_map).difference(no_iscsi_status))}")
 
     comp_patches = []
     for xname, comp in worker_comp_map.items():
-        patch = { "error_count": 0, "id": xname }
-        comp_patches.append(patch)
         if xname in no_iscsi_status:
             continue
-        patch["state"] = []
+        state = []
         for layer in comp["state"]:
             if layer["playbook"] == ISCSI_PLAYBOOK:
                 continue
             # Cannot patch the last_updated field
             del layer["last_updated"]
-            patch["state"].append(layer)
+            state.append(layer)
+        comp_patches.append({ "error_count": 0, "id": xname, "state": state })
 
     print("Updating CFS components...")
     update_components_by_list(comp_patches)
--- a/scripts/operations/configuration/refresh_worker_iscsi_config.py
+++ b/scripts/operations/configuration/refresh_worker_iscsi_config.py
@@ -81,7 +81,7 @@ def main() -> None:
 
     # First check to make sure we got back one component for every worker
     worker_cfs_comp_ids = sorted([ comp["id"] for comp in worker_cfs_comps ])
-    missing_xnames = set(worker_ncn_xnames).difference_update(worker_cfs_comp_ids)
+    missing_xnames = set(worker_ncn_xnames).difference(worker_cfs_comp_ids)
     if missing_xnames:
         raise ScriptException(
             f"No CFS component found for following worker NCNs: {sorted(missing_xnames)}")
--- a/scripts/operations/configuration/refresh_worker_iscsi_config.py
+++ b/scripts/operations/configuration/refresh_worker_iscsi_config.py
@@ -109,17 +109,22 @@ def main() -> None:
           if cfs_comp_has_no_iscsi_status(comp)
         ]
     )
+    have_iscsi_status = sorted(set(worker_comp_map).difference(no_iscsi_status))
+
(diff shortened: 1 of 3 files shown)


scripts/operations/configuration/refresh_worker_iscsi_config.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Warn if any workers have no iSCSI layer in their status
+    no_iscsi_status = sorted(
+        [ xname for xname, comp in worker_comp_map.items()
+          if cfs_comp_has_no_iscsi_status(comp)
+        ]
+    )
+
+    if no_iscsi_status:
+        print("WARNING: The following workers have no iSCSI layer in their CFS "
```
kumarrahul04, line 119: resetting the error count will trigger CFS operation. If components do not have iSCSI layer do we need cfs configuration triggered?

mharding-hpe: Resetting the error count will only trigger the CFS operation if it had previously failed. If it successfully configured previously, it will have no effect.

I debated which way to go on that one. In most cases it won't make a difference. I'm happy changing it to just skip those nodes and keep the warning.

kumarrahul04: I think the script is about updating the iSCSI layer. we should not induce CFS configuration for other nodes as it can happen in certain situation.

mharding-hpe: Changed.

kumarrahul04 (approved):

dborman-hpe:
Just to clarify my understanding - the clearing of the CFS iscsi state is for all worker nodes, so that when CFS is re-run it will re-run iscsi for all worker nodes, which will then clear/set the label on all the worker nodes based on the HSM list of iscsi nodes.  Correct?

mharding-hpe:
> Just to clarify my understanding - the clearing of the CFS iscsi state is for all worker nodes, so that when CFS is re-run it will re-run iscsi for all worker nodes, which will then clear/set the label on all the worker nodes based on the HSM list of iscsi nodes. Correct?

It is for all worker nodes, yes. And when they re-run the iSCSI playbook, it will run all of the roles, including the one which updates the labels for each node.

Well, to be more precise, it is for every worker node that HAS an iSCSI state. If a worker node does not have one, that either means CFS has never run on it, or it is using a desired configuration that does not include an iSCSI layer. For those worker nodes, it does not clear the iSCSI state because, well, there's nothing to clear.

In those cases, it will only be running the iSCSI playbook on some of the worker nodes (the one whose state it was able to clear). With the current code in `csm-config`, this will mean that for the uncleared worker nodes, their labels will not be changed. The current code will also fail to update labels for any worker nodes that are down or unreachable, even if we did clear their state above.

If we get the [CASMPET-7624/CASMPET-7625 changes](https://github.com/Cray-HPE/csm-config/pull/385) (waiting on a review from @aasha-hpe ), then none of those will be issues any longer, because that changes things so that the labels for all workers (up or down) get updated if even a single worker node runs the play.

dborman-hpe (approved):
LGTM

aasha-hpe (approved):

PR #6132, opened 2025-07-17
CASMPET-7619/CASMPET-7620/CASMPET-7621/CASMPET-7630/CASMPET-7632: Create/fix procedures for changing iSCSI workers

aasha-hpe:
# Description
Documentation on selective node personalization or limit/not limit the worker nodes configuration for iSCSI SBPS.

Resolves:
* CASMPET-7619
* CASMPET-7620
* CASMPET-7621
* CASMPET-7630
* CASMPET-7632

Relates to:
* CASMPET-7616
* https://github.com/Cray-HPE/csm-config/pull/381

- [x] If I added any command snippets, the steps they belong to follow the prompt conventions (see [example][1]).
- [x] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].
- [x] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

PR #6134, opened 2025-07-17
CASMPET-7620: DOCS: Update CSM install docs with iSCSI worker step

aasha-hpe:
# Description
This is to update CSM install docs with the iSCSI worker selective node personalization as per https://jira-pro.it.hpe.com:8443/browse/CASMPET-7619. Referencing this doc in CSM 1.7.0 installation doc. 

Relates to:
CASMPET-7619
https://github.com/Cray-HPE/docs-csm/pull/6132

### DIFF
--- a/install/configure_administrative_access.md
+++ b/install/configure_administrative_access.md
@@ -19,7 +19,8 @@ and BMC/controller passwords.
 1. [Set up passwordless SSH](#5-set-up-passwordless-ssh)
 1. [Configure the root password and SSH keys in Vault](#6-configure-the-root-password-and-ssh-keys-in-vault)
 1. [Add switch admin password to Vault](#7-add-switch-admin-password-to-vault)
-1. [Configure management nodes with CFS](#8-configure-management-nodes-with-cfs)
+1. [iSCSI SBPS configuration](#8-iscsi-sbps-configuration)
+1. [Configure management nodes with CFS](#9-configure-management-nodes-with-cfs)
 1. [Proceed to next topic](#9-proceed-to-next-topic)
 
 > **`NOTE`** The procedures in this section of installation documentation are intended to be done in order, even though the topics are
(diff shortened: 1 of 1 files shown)


install/configure_administrative_access.md
```diff
@@ -159,7 +160,23 @@ Password read from Vault matches what was written
...
-## 8. Configure management nodes with CFS
+## 8. iSCSI SBPS configuration
+
+In CSM 1.6.0, all the worker nodes were configured as iSCSI SBPS targets by default. In CSM 1.7.0, selective
+node personalization is supported where if user/admin wants to limit some worker nodes to be configured as
+iSCSI targets, then it requires to create an HSM group name `iscsi_worker` and add the worker node xnames
+of the nodes which are required to be configured as iSCSI targets to this group. For the steps/procedure for
+the same, see [HSM groups iSCSI](https://github.com/Cray-HPE/docs-csm/blob/release/1.7/operations/iscsi_sbps/HSM_groups_iscsi.md)
+
+This has to be done before "Configure management nodes with CFS" during CSM installation. If this is not done,
```
mharding-hpe: Actually, this is not right, I don't think. Isn't SAT what adds the iSCSI layer into the CFS configuration? The `Configure management nodes with CFS` step will not include that layer. It only includes `ncn_nodes.yml` and `ncn-initrd.yml`.

My first thought was that we could modify the script used in that step so that it would include the iSCSI layer as well, but then I remember that we talked about something related to this on one of our sync calls. I had mentioned that now that we are running this on all of the worker nodes, we should just combine it with the `ncn_nodes.yml` playbook, rather than have it be its own playbook. And someone stated that they weren't sure if the playbook would work properly if Slingshot had not been configured. They said all of the testing had been done assuming Slingshot would be configured, so they didn't want to change it. In that case, we also wouldn't want to add this layer to the script that runs in the `Configure management nodes with CFS` step.

Probably this should be re-worded to say something like: 
```
Later in the install process, SAT bootprep adds the iSCSI configuration layer to the NCN CFS configurations. If the iSCSI HSM group does not exist when that CFS configuration is applied, then all of the worker nodes will be enabled as iSCSI targets.
```

mharding-hpe:
These changes are being made with [a different PR](https://github.com/Cray-HPE/docs-csm/pull/6132) instead, so I am closing this one

aasha-hpe:
The changes are part of https://github.com/Cray-HPE/docs-csm/pull/6132

PR #6142, opened 2025-07-23
CASMPET-7632: DOCS: Remove/move iSCSI_SBPS_Node_Personalization.md

aasha-hpe:
# Description
This is to remove https://github.com/Cray-HPE/docs-csm/blob/release/1.7/operations/configuration_management/iSCSI_SBPS_Node_Personalization.md as this is not required with the introduction of selective worker node personalization. The post personlisation verification steps is added as separate document under 'iscsi_sbps' folder as 'Node_Personalization_Verification.md'

Relates to:
CASMPET-7619
https://github.com/Cray-HPE/docs-csm/pull/6132

# Checklist

- [NA] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].


operations/iscsi_sbps/Node_Personalization_Verification.md
```diff
@@ -1,238 +1,4 @@
...
-    ```
-
-    Example of output if CFS was successful:
-
-    ```text
-    All playbooks completed successfully
-    ```
-
-### Post-personalization verification
+# Post-personalization Verification
```
mharding-hpe, line 1: The top-level header and file name should match.

operations/iscsi_sbps/Node_Personalization_Verification.md
```diff
@@ -1,238 +1,4 @@
...
 
 (`ncn-w#`) Run these checks on each worker node, in order to verify that it was configured correctly.
```
mharding-hpe, line 3: Context needs to be added here. What configuration is being verified on the worker node?

You don't need to say a lot, because obviously this page is in the iscsi_sbps folder, so people could reasonably deduce that this is related to that. But it should be explicitly mentioned here as well, just to make it clear.

operations/iscsi_sbps/Node_Personalization_Verification.md
mharding-hpe, line 1: There are no links to this new page. Any time a new page is added, we need to make sure that links are created to it in the places where users will find helpful.

At a minimum this should probably be linked from the main iSCSI page (iscsi_sbps), the operations index (`operations/README.md`), and possibly the Managing selective node personalization page (any time after the CFS script is run, it should probably then have a pointer to this page, showing them how they can verify that the configuration was completed successfully).

operations/iscsi_sbps/Node_Personalization_Verification.md
mharding-hpe, line 3: If they run this on every worker node, won't they see issues for the workers that are not enabled as iSCSI targets? Or are some of these checks that iSCSI is configured (and therefore would apply to all worker nodes) and some of these are checks that the worker is enabled as an iSCSI target (which may not work on all worker nodes)?

mharding-hpe (changes requested):

mharding-hpe:
At the moment, this change has been combined into https://github.com/Cray-HPE/docs-csm/pull/6132 .
In light of that, I'll close this PR.
If @aasha-hpe later decides she'd rather make those changes separately, this PR can of course be reopened.

PR #6149, opened 2025-07-25
CASMNET-2360 - Exclude SMA kafka network policies from Cilium migration

spillerc-hpe:
# Description

The Cilium migration workflow already excludes the `cray-shared-kafka` NetworkPolicy resources from the backup taken during the `stash-netpol` step as they are managed by the operator and are recreated as soon as they are deleted.

The SMA Kafka cluster policies require the same treatment.
```
ncn-m001:~ # kubectl -n sma get netpol
NAME                               POD-SELECTOR                                                                               AGE
cluster-entity-operator            strimzi.io/cluster=cluster,strimzi.io/kind=Kafka,strimzi.io/name=cluster-entity-operator   16h
cluster-kafka-exporter             strimzi.io/cluster=cluster,strimzi.io/kind=Kafka,strimzi.io/name=cluster-kafka-exporter    16h
cluster-network-policy-kafka       strimzi.io/cluster=cluster,strimzi.io/kind=Kafka,strimzi.io/name=cluster-kafka             16h
cluster-network-policy-zookeeper   strimzi.io/cluster=cluster,strimzi.io/kind=Kafka,strimzi.io/name=cluster-zookeeper         17h
```
This change has been tested on wasp.

# Checklist

### DIFF
--- a/workflows/templates/cilium.deploy.yaml
+++ b/workflows/templates/cilium.deploy.yaml
@@ -135,8 +135,8 @@ spec:
                   rm -f /tmp/netpol.yaml
                   for i in $(kubectl get netpol -A --no-headers | awk '{print $1}' | sort -u );do
                     echo "--- namespace ${i} ---"
-                    # Not stashing network policies for cray-shared-kafka because they get recreated by the operator
-                    for j in $(kubectl -n ${i} get netpol --no-headers | grep -v cray-shared-kafka | awk '{print $1}');do
+                    # Not stashing network policies for cray-shared-kafka or SMA kafka because they get recreated by the operator
+                    for j in $(kubectl -n ${i} get netpol --no-headers | egrep -v "cray-shared-kafka|cluster" | awk '{print $1}');do
                       echo "Saving off network policy ${j}"
                       echo "---" >> /tmp/netpol.yaml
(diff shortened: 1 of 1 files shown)


jpdavis-prof (approved):

studenym-hpe (approved):

mharding-hpe (approved):
@spillerc-hpe you know more about this than me, but could this also be done just by filtering out the `sma` namespace itself? Or is the concern that `sma` may have some netpolicies other than the ones we're trying to exclude? On wasp that doesn't seem to be the case, but I realize that's just one system. 

I was also slightly concerned about filtering out a vague word like `cluster` across the board. I see on wasp this doesn't pull in anything unwanted, but it seems like the sort of word that could end up in other netpol names or pod selectors (since these filters are applying to the total `kubectl` output, not just the netpol names). Again, you know a lot more about this than me.

I'm approving, but did want to raise these concerns, since it seems possible to do a more surgical approach to avoid filtering things accidentally. But I realize that the current choice is the minimal change, just expanding the current filter, so I do see the advantages

spillerc-hpe:
@mitcharf it actually doesn't matter if we catch more in this case, we want to delete the all the network policies (they're restored later). For these two specific cases we can't because the Kafka operator helpfully recreates them if you remove them.

PR #6179, opened 2025-08-01
 CASMTRIAGE-8555: Imports rack-resiliency settings during upgrade

arka-pramanik-hpe:
Adds rack-resiliency configuration import from upgrade customizations file since this service did not have configurable customization prior to version 1.7.

Resolves [CASMTRIAGE-8555](https://jira-pro.it.hpe.com:8443/browse/CASMTRIAGE-8555)
Resolves [CASMTRIAGE-8547](https://jira-pro.it.hpe.com:8443/browse/CASMTRIAGE-8547)

# Description

# Testing

`odin`

# Checklist

### DIFF
--- a/upgrade/scripts/upgrade/util/update-customizations.sh
+++ b/upgrade/scripts/upgrade/util/update-customizations.sh
@@ -125,8 +125,8 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
   ' "$c"
 fi
 
-# rack-resiliency did not have configurable customization prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
-if [ "$(yq4 eval '.spec.kubernetes.services[\"rack-resiliency\"]' "$c")" == "null" ]; then
+# rack-resiliency did not exist prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
+if [ "$(yq4 eval '.spec.kubernetes.services.rack-resiliency' "$c")" == null ]; then
   yq4 -i eval ".spec.kubernetes.services[\"rack-resiliency\"] += (load(\"${upgrade_customizations}\") | .spec.kubernetes.services[\"rack-resiliency\"])" "$c"
 fi
 
--- a/upgrade/scripts/upgrade/util/update-customizations.sh
+++ b/upgrade/scripts/upgrade/util/update-customizations.sh
@@ -125,6 +125,9 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
   ' "$c"
 fi
 
+# rack-resiliency did not have configurable customization prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
+yq4 -i eval ".spec.kubernetes.services[\"rack-resiliency\"] += (load(\"${upgrade_customizations}\") | .spec.kubernetes.services[\"rack-resiliency\"])" "$c"
+
 # Add entries for Cilium Hubble observability GUI (CASMNET-2194)
 yq4 -i eval '.spec.kubernetes.services.["cray-hubble"].externalHostname = "hubble.cmn.{{ network.dns.external }}"' "$c"
 yq4 -i eval ".spec.proxiedWebAppExternalHostnames.customerManagement |= (. + [\"{{ kubernetes.services['cray-hubble'].externalHostname }}\"] | unique)" "$c"
--- a/upgrade/scripts/upgrade/util/update-customizations.sh
+++ b/upgrade/scripts/upgrade/util/update-customizations.sh
@@ -126,7 +126,9 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
 fi
 
 # rack-resiliency did not have configurable customization prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
-yq4 -i eval ".spec.kubernetes.services[\"rack-resiliency\"] += (load(\"${upgrade_customizations}\") | .spec.kubernetes.services[\"rack-resiliency\"])" "$c"
+if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "$c")" == "null" ]; then
+  yq4 -i eval ".spec.kubernetes.services[\"rack-resiliency\"] += (load(\"${upgrade_customizations}\") | .spec.kubernetes.services[\"rack-resiliency\"])" "$c"
+fi
 
 # Add entries for Cilium Hubble observability GUI (CASMNET-2194)
 yq4 -i eval '.spec.kubernetes.services.["cray-hubble"].externalHostname = "hubble.cmn.{{ network.dns.external }}"' "$c"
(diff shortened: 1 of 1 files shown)


upgrade/scripts/upgrade/util/update-customizations.sh
```diff
@@ -125,6 +125,9 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
   ' "$c"
 fi
 
+# rack-resiliency did not have configurable customization prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
```
don-bahls-hpe: I would think you'd want to check if the rack-resiliency stanza existed before updating it (similar to what is being done on line 106).   Did you test this change with an existing "rack-resiliency" stanza in the yaml file?

mharding-hpe: I agree, but then that makes me think we'd want to do the same for the Kyverno policy setting on 116. Unless that yq += is a patch that preserves existing fields and only sets ones that don't already exist. But I definitely agree with Don -- you need to test this to make sure it won't overwrite the values if they already exist

mharding-hpe: I also think that the piping in line 123 may be trying to accomplish exactly this (ignoring the del part, which wouldn't apply here). It looks to me like it's starting with the values from the update customizations.yaml file and then patching the values from the existing customizations.yaml file. I'm guessing that if they don't exist in the current customizations.yaml, then that patch would just be a no-op

don-bahls-hpe: I'm not totally following how the sub keys would get added.   Do these keys get added?
```
      rack-resiliency:
        enabled: false
        k8s_zone_prefix: ""
        ceph_zone_prefix: ""
```

ravikanth-nalla-hpe (approved):
LGTM

upgrade/scripts/upgrade/util/update-customizations.sh
ravikanth-nalla-hpe: @mharding-hpe @don-bahls-hpe Since this is the first release of RR we want to copy the new RR section from the new customization file to the old customization file during upgrade. However, if somebody is upgrading from 1.7.0 to 1.7.0 (during testing) make sure we don't loose customization we are adding this if condition:

```
if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "$c")" == "null" ]; then
  yq4 -i eval ".spec.kubernetes.services[\"rack-resiliency\"] += (load(\"${upgrade_customizations}\") | .spec.kubernetes.services[\"rack-resiliency\"])" "$c"
fi

``` 
I hope with this you can approve.

keshav-06-hpe: @don-bahls-hpe , @mitcharf 
Below testcases have been run:
```
Case:1 
Old customizations has no rr and new has customizations(1.7) has RR disabled

output after running update-customizations.sh:
ncn-m001:~/arka/custom # sh update-customizations.sh -i old-cust.yaml customizations.yaml 

old file will have now:
   rack-resiliency:
        enabled: false
        # If k8s_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        k8s_zone_prefix: ""
        # If ceph_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        ceph_zone_prefix: ""

Case:2
Old customizations has RR disabled and new has customizations(1.7) has RR disabled
No changes observed after running update-customizations.sh: 
    rack-resiliency:
        enabled: false
        # If k8s_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        k8s_zone_prefix: ""
        # If ceph_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        ceph_zone_prefix: ""

Case:3 
Old customizations has RR enabled but new customizations has RR disabled

entry in new custom file for RR looks like:
    rack-resiliency:
        enabled: false
      # If k8s_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        k8s_zone_prefix: "cscs-rack"
      # If ceph_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        ceph_zone_prefix: "cscs-rack"

old customizations will not be updated by the values in new customizations for RR:
    rack-resiliency:
        enabled: true
        # If k8s_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        k8s_zone_prefix: "cscs-rack"
        # If ceph_zone_prefix is not updated, by default it will be having the rack-id as the zone name
        ceph_zone_prefix: ""

Summary: All subkeys are added and modified during this process.
```

rajeshranganathan-hpe (approved):

upgrade/scripts/upgrade/util/update-customizations.sh
don-bahls-hpe: I would think case 3 should leave the customizations.yaml alone.

don-bahls-hpe (approved):

upgrade/scripts/upgrade/util/update-customizations.sh
keshav-06-hpe: @don-bahls-hpe, Yes, it is leaving the old `customizations.yaml` as it is

upgrade/scripts/upgrade/util/update-customizations.sh
```diff
@@ -125,6 +125,11 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
...
+if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "$c")" == "null" ]; then
```
mharding-hpe: Why are you checking for image-verification policy here?

mharding-hpe (changes requested):

upgrade/scripts/upgrade/util/update-customizations.sh
keshav-06-hpe: It has been updated. Please re-review.

upgrade/scripts/upgrade/util/update-customizations.sh
```diff
@@ -125,6 +125,11 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
   ' "$c"
 fi
 
+# rack-resiliency did not have configurable customization prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
+if [ "$(yq4 eval '.spec.kubernetes.services[\"rack-resiliency\"]' "$c")" == "null" ]; then
```
mharding-hpe: Is there a reason not to simplify this (which also would make it consistent with how the check on line 106 does it)?
```suggestion
if [ "$(yq4 eval '.spec.kubernetes.services.rack-resiliency' "$c")" == null ]; then
```

upgrade/scripts/upgrade/util/update-customizations.sh
```diff
@@ -125,6 +125,11 @@ if [ "$(yq4 eval '.spec.kubernetes.services."image-verification-policy"' "${upgr
   ' "$c"
 fi
 
+# rack-resiliency did not have configurable customization prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
```
mharding-hpe:
```suggestion
# rack-resiliency did not exist prior to 1.7. Import rack-resiliency settings from upgrade customizations file during upgrade.
```

mharding-hpe (approved):

PR #6183, opened 2025-08-04
CASM-5499: Add conditional deployment and enablement of RRS (Rack Res…

ravikanth-nalla-hpe:
…iliency Service)

- Support documentation for enabling RR and adding zone prefixes in customizations.yaml file during install

# Description
- During install, we need to support documentation for enabling Rack Resiliency and adding zone (Kubernetes and Ceph) prefixes in customizations.yaml file.

### DIFF
--- a/install/prepare_site_init.md
+++ b/install/prepare_site_init.md
@@ -10,6 +10,7 @@ directory which contains important customizations for various products.
     1. [Customize DNS Configuration](#customize-dns-configuration)
     1. [Configure Prometheus SNMP Exporter](#configure-prometheus-snmp-exporter)
     1. [Configure Max Istio Gateway Pods](#configure-max-istio-gateway-pods)
+    1. [Enable Rack Resiliency](#enable-rack-resiliency)
 1. [Encrypt secrets](#4-encrypt-secrets)
 1. [Customer-Specific Customizations](#5-customer-specific-customizations)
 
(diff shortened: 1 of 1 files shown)


mharding-hpe (changes requested):
This is currently failing the linter. Also, is there a reason this is only covering the install path and not the upgrade path?

install/prepare_site_init.md
```diff
@@ -452,7 +453,28 @@ the system size.
         "$min_amount"
     done
     ```
+### Enable Rack Resiliency
 
+This section can be skipped if Rack Resiliency feature is not required.
```
mharding-hpe, line 458: This cannot be included without giving some information about rack resiliency. Otherwise there is no intelligent way for them to make the decision. This is what the "minimal PR" I sent you a message about includes.

install/prepare_site_init.md
```diff
@@ -452,7 +453,28 @@ the system size.
...
+
+1. (`pit#`) Update the customizations file.
+
+    ```bash
+      yq write -i ${SITE_INIT}/customizations.yaml \
+        'spec.kubernetes.services.rack-resiliency.enabled' "true"
+    ```
+1. (`pit#`) If site specific prefix for Kubernetes and `Ceph` zones is required:
```
mharding-hpe, line 466: This also needs to provide some way for an admin to make this decision. How do they know if it is required? Again, this is covered by the "minimal PR" I outlined for you in Slack.

install/prepare_site_init.md
```diff
@@ -452,7 +453,28 @@ the system size.
...
+
+   For example:
+   
+   ```bash
+      yq write -i ${SITE_INIT}/customizations.yaml \
+        'spec.kubernetes.services.rack-resiliency.k8s_zone_prefix' "prefix-string"
```
mharding-hpe, line 472: This example makes it seem like the prefix for both is expected to match. Is that a requirement?

ravikanth-nalla-hpe: Nope. It can be any string. It is just an example.

PR #6236, opened 2025-08-12
CASM-5679: Update install and upgrade procedures for RR

ravikanth-nalla-hpe:
# Description
[CASM-5679](https://jira-pro.it.hpe.com:8443/browse/CASM-5679): RR playbook disruptive to system

The system is unacceptably disrupted when the RR kyverno policy playbook does restarts of  critical services all at once. When this happened on the first master node in one of the system, it caused delays in settling down (not quickly).This delays may cause issues. 

- The restart of the critical services has been now moved away from RR ansible play books and added to the upgrade k8s scripts. Here critical services are rollout restarted with timeout (wait for completion of the previous critical service rollout restart) and also they are restaretd only once and not at every upgrade. 
- Additional conditional checks also has been added to rollout restart these critical services only when RR is enabled and Kyverno policy is applied/ present otherwies skip to restart.
- This will also update "rollout_complete" flag in dynamic ConfigMap in order for RRS to wait before continuing to init RRS.

# Checklist


upgrade/scripts/k8s/upgrade_k8s_steps.sh
```diff
@@ -119,5 +119,18 @@ else
   fi
 fi
 
+if [[ -f "$DONE_DIR/rr_cs_rollout_restart.done" ]]; then
```
mharding-hpe, line 122: Adding this here will address the CSM upgrade case, but this also needs to happen on CSM fresh installs.

upgrade/scripts/k8s/cm_static.yml
mharding-hpe, line 1: The static configmap is being added to csm-config with [this PR](https://github.com/Cray-HPE/csm-config/pull/400), so it should not be duplicated here

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+    yq_cmd = ["yq", "r", decoded_yaml, key_path]
+    try:
+        result = subprocess.run(yq_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, check=True)
+        if result.returncode != 0:
+            raise ValueError(f"Error fetching site-init secret: {result.stderr}")
+        return json.loads(result.stdout)
+    except Exception as e:
+        return {"error": str(e)}
+
+    # Extract and clean the output
```
mharding-hpe: Should add a comment here similar to what is in the corresponding RR code, explaining that this needs to do the same test as Ansible `bool`, so people know why the check is like this

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Extract and decode the base64 data
+    encoded_yaml = secret_data["data"]["customizations.yaml"]
+    decoded_yaml = base64.b64decode(encoded_yaml).decode("utf-8")
+
+    # Define the key path
+    key_path = "spec.kubernetes.services.rack-resiliency.enabled"
+
+    # Write the yaml output to a file
+    output_file = "/tmp/customization.yaml"
```
mharding-hpe: Hard coding the temporary file name and path like this is not ideal, especially since `/tmp` is a common place where someone in the interactive shell may create a file with this name. I'd recommend using the Python `tempfile` module.

But in this case, why are you writing this to a file only so you can run `yq` on that file? Why not just use the Python `yaml` module? I know it is not a built-in Python module, but `yq` isn't built into SLES either, so either way you're relying on something being present. The `yaml` module is installed on the NCNs, so in this case it would simplify your life considerably just to use that.

And I see that you are already using the `yaml` module elsewhere in the file, so you're already dependent on it.

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+    with open(output_file, "w") as f:
+        f.write(decoded_yaml)
+
+    # Define the key path
+    output_file = "/tmp/customization.yaml"
+    key_path = "spec.kubernetes.services.rack-resiliency.enabled"
```
mharding-hpe: Aren't these already being done on lines 165 and 168?

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+        get_command = [
+            "kubectl", "get", resource_type, name,
+            "-n", namespace, "-o", "json"
+        ]
+
+        try:
+            result = subprocess.run(get_command, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
+            resource_json = json.loads(result.stdout)
+
+            # Check to if 'rrflag' is set for critical service
```
mharding-hpe:
```suggestion
            # Check if 'rrflag' is set for critical service
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+"""
+
+import yaml
+import json
+import subprocess
+import sys
+import base64
+import os
+
+def load_configmap(name, namespace):
```
mharding-hpe:
```suggestion
def load_configmap(name: str, namespace: str) -> dict:
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+            universal_newlines=True
+        )
+        return json.loads(result.stdout)
+    except subprocess.CalledProcessError as e:
+        print(f"Failed to fetch ConfigMap '{name}' from namespace '{namespace}'")
+        print(f"Error: {e.stderr}")
+        sys.exit(1)
+
+
+def rollout_restart_critical_services(critical_services):
```
mharding-hpe:
```suggestion
class ServiceDetails(TypedDict):
    type: str
    namespace: str

def rollout_restart_critical_services(critical_services: Dict[str, ServiceDetails]) -> Literal[0, 1]:
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+"""
+Do rollout restart of the critical services defined in RR static ConfigMap.
+"""
+
+import yaml
+import json
+import subprocess
+import sys
+import base64
+import os
```
mharding-hpe:
```suggestion
import json
import subprocess
import sys
import base64
import os
from typing import Dict, List

from typing_extensions import Literal, TypedDict
import yaml
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+        except subprocess.CalledProcessError as e:
+            if "not found" in e.stderr.lower():
+                print(f"Skipping {resource_type}/{name}: resource not found in namespace {namespace}")
+                continue
+            print(f"Failed to restart {resource_type}/{name} in namespace {namespace}")
+            print(f"Error: {e.stderr}")
+            return 1
+    return 0
+
+def set_rollout_complete(configmap_name, namespace):
```
mharding-hpe:
```suggestion
def set_rollout_complete(configmap_name: str, namespace: str) -> None:
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+    try:
+        subprocess.run(command, check=True)
+        print(f"Set rollout_complete=true in ConfigMap '{configmap_name}'")
+    except subprocess.CalledProcessError as e:
+        print(f"Failed to patch ConfigMap '{configmap_name}'")
+        print(f"Error: {e}")
+        sys.exit(1)
+
+
+def rr_enabled():
```
mharding-hpe, line 176:
```suggestion
def rr_enabled() -> bool:
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+    # Extract and clean the output
+    enabled = result.stdout.strip()
+    if any(enabled is tvalue for tvalue in [1, 1.0, True]):
+        return True
+    if not isinstance(enabled, str):
+        return False
+    return enabled.lower() in {"true", "t", "yes", "y", "on", "1"}
+
+
+def is_cluster_policy_applied(policy_name):
```
mharding-hpe:
```suggestion
def is_cluster_policy_applied(policy_name: str) -> bool:
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+            check=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL
+        )
+        return True
+    except subprocess.CalledProcessError:
+        return False
+
+
+def main():
```
mharding-hpe:
```suggestion
def main() -> None:
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+    Main function to execute the rollout restart of critical services.
+    """
+    # Check if RR is enabled and if the cluster policy is applied
+    if not (rr_enabled() and is_cluster_policy_applied("insert-labels-topology-constraints")):
+        print("Either Rack Resiliency is not enabled or ClusterPolicy 'insert-labels-topology-constraints' not found. Skipping restart.")
+        sys.exit(0)
```
mharding-hpe, line 265: If RR is enabled but the policy is not applied, shouldn't that be an error? Since this is the only time this rollout script is being run, wouldn't we want it to fail in that case?

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+            resource_json = json.loads(result.stdout)
+
+            # Check to if 'rrflag' is set for critical service
+            labels = resource_json.get("spec", {}).get("template", {}).get("metadata", {}).get("labels", {})
+            if "rrflag" not in labels:
+                print(f"Skipping {resource_type}/{name}: 'rrflag' label is not set for {resource_type}/{name} in namespace {namespace}")
+                continue
+
+            # Restart the critical service
+            restart_command = ["kubectl", "rollout", "restart", f"{resource_type}/{name}", "-n", namespace]
```
mharding-hpe: If this IUF step fails and this step has to be re-run, I don't see any logic in here to make sure we aren't just restarting the same services over and over. Given how many it has to restart, we should try to avoid doing it unnecessarily.

It seems like it ought to be possible to compare the oldest pod in a deployment and the timestamp when the Kyverno policy was last applied. If all pods in a deployment/statefulset were started after the Kyverno policy was in place, then no need to restart it.

Or maybe even easier, it looks like there is a metadata annotation for `kubectl.kubernetes.io/restartedAt`, so you could compare that to the Kyverno policy apply timestamp.

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,242 @@
...
+            status_command = ["kubectl", "rollout", "status", resource_type, name, "-n", namespace, "--timeout=3m"]
+
+            subprocess.run(restart_command, check=True,stdout=subprocess.PIPE,stderr=subprocess.PIPE, universal_newlines=True)
+            subprocess.run(status_command, check=True,stdout=subprocess.PIPE,stderr=subprocess.PIPE, universal_newlines=True)
+            print(f"Restarted {resource_type}/{name} in namespace {namespace}")
+        except subprocess.CalledProcessError as e:
+            if "not found" in e.stderr.lower():
```
mharding-hpe: It would be better to use two different try/except blocks. Or really, three different ones, one for each kubectl command. That would mean you wouldn't need to check `stderr` for a specific string like this, and it would also mean that you could distinguish between failing to restart the deployment versus the restart status failing or timing out.

Plus, it's generally best practice to make try/excepts small, covering only one thing.

mharding-hpe (changes requested):

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+Do rollout restart of the critical services defined in RR static ConfigMap.
+"""
+
+import json
+import subprocess
+import sys
+import base64
+import os
+from typing import Dict, List
```
mharding-hpe:
```suggestion
from typing import Dict
```

upgrade/scripts/k8s/cm_static.yml
ravikanth-nalla-hpe: So, we want to consider this ConfigMap it is already configured by ansible plays by this time? Can me remove applying it here and just read the config?

upgrade/scripts/k8s/rr_critical_service_restart.py
ravikanth-nalla-hpe: What exact comment we are talking about?

rajeshranganathan-hpe: we will do it sir

upgrade/scripts/k8s/rr_critical_service_restart.py
ravikanth-nalla-hpe: RR label "rrflag" is added to the deployment during rollout restart. This is verified  before restart and the service is skipped.

upgrade/scripts/k8s/rr_critical_service_restart.py
mharding-hpe: https://github.com/Cray-HPE/cray-rrs/blob/b51fb347c16e825c3a7eaf6cf15ad4c827acc858/src/rrs/init/wait.py#L134-L139

upgrade/scripts/k8s/rr_critical_service_restart.py
mharding-hpe: Ah, I did not realize that. Or, if I did realize that at one time, I have forgotten it since then. :) That resolves this comment.

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,278 @@
...
+    except subprocess.CalledProcessError:
+        return False
+
+def main():
+    """
+    Main function to execute the rollout restart of critical services.
+    """
+    # Check if RR is enabled and cluster policy is applied
+    if rr_enabled() and not is_cluster_policy_applied("insert-labels-topology-constraints"):
+        print("Rack Resiliency is enabled but ClusterPolicy 'insert-labels-topology-constraints' is not applied. Skipping restart.")
```
mharding-hpe:
```suggestion
        sys.stderr.write("ERROR: Rack Resiliency is enabled but ClusterPolicy 'insert-labels-topology-constraints' is not applied. Skipping restart.\n")
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,278 @@
...
+        sys.exit(1)
+    # Check if RR is not enabled and cluster policy is not applied
+    elif not rr_enabled() and not is_cluster_policy_applied("insert-labels-topology-constraints"):
```
mharding-hpe:
```suggestion
    if not rr_enabled() and not is_cluster_policy_applied("insert-labels-topology-constraints"):
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,278 @@
...
+        print("Rack Resiliency is not enabled and ClusterPolicy 'insert-labels-topology-constraints' is not applied.Skipping restart.")
+        sys.exit(1)
+    # Check if RR is not enabled but cluster policy is applied
+    elif not rr_enabled() and is_cluster_policy_applied("insert-labels-topology-constraints"):
```
mharding-hpe:
```suggestion
    if not rr_enabled() and is_cluster_policy_applied("insert-labels-topology-constraints"):
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,278 @@
...
+        print("Rack Resiliency is not enabled  but ClusterPolicy 'insert-labels-topology-constraints' is applied. Skipping restart.")
```
mharding-hpe:
```suggestion
        sys.stderr.write("ERROR: Rack Resiliency is not enabled  but ClusterPolicy 'insert-labels-topology-constraints' is applied. Skipping restart.\n")
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,278 @@
...
+        print(f"Missing expected key in ConfigMap: {e}")
+        sys.exit(1)
+
+    if rollout_restart_critical_services(critical_services) == 0:
+        print(f"RR critical services rollout restart successful.")
+        # Set "rollout_complete" to "true" in RR dynamic ConfigMap
+        set_rollout_complete("rrs-mon-dynamic", "rack-resiliency")
+    else:
+        print(f"RR critical services rollout restart failed.")
+        sys.exit(1)
```
mharding-hpe:
```suggestion
    if rollout_restart_critical_services(critical_services) != 0:
        sys.stderr.write(f"RR critical services rollout restart failed.\n")
        sys.exit(1)
    print(f"RR critical services rollout restart successful.")
    # Set "rollout_complete" to "true" in RR dynamic ConfigMap
    set_rollout_complete("rrs-mon-dynamic", "rack-resiliency")
```

mharding-hpe:
Had a call today with the RR development team. Based on that, there are some planned additions to what this PR is doing, which will probably resolve the remaining comments.

upgrade/scripts/k8s/upgrade_k8s_steps.sh
mharding-hpe: We discussed on the call how to address this

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,277 @@
...
+Do rollout restart of the critical services defined in RR static ConfigMap.
+"""
+
+import yaml
+import json
+import subprocess
+import sys
+import base64
+import os
+import tempfile
```
mharding-hpe:
```suggestion
import json
import subprocess
import sys
import base64
import os

import yaml
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,282 @@
...
+        result = subprocess.run(
+            ["kubectl", "get", "configmap", name, "-n", namespace, "-o", "json"],
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True
+        )
+        return json.loads(result.stdout)
+    except subprocess.CalledProcessError as e:
+        print_stderr(f"Failed to fetch ConfigMap '{name}' from namespace '{namespace}'")
```
mharding-hpe:
```suggestion
        print_err(f"Failed to fetch ConfigMap '{name}' from namespace '{namespace}'")
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    sys.stderr.write(f"{msg}\n")
+    sys.stderr.flush()
+
+def err_exit(msg: str) -> NoReturn:
+    """
+    Prepends "ERROR: " to the message and then prints it to stderr.
+    Then exits the script with rc 1
+    """
+    print_stderr(f"ERROR: {msg}")
```
mharding-hpe:
```suggestion
    print_err(f"ERROR: {msg}")
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,282 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True
+        )
+        return json.loads(result.stdout)
+    except subprocess.CalledProcessError as e:
+        print_err(f"Failed to fetch ConfigMap '{name}' from namespace '{namespace}'")
+        err_exit(e.stderr)
+
+class ServiceDetails(TypedDict):
```
mharding-hpe, line 80:
```suggestion
class ServiceDetails(TypedDict):
    """
    The type (e.g. "Deployment", "StatefulSet") and namespace (e.g. "services")
    for a critical service in Kubernetes
    """
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,286 @@
...
+    try:
+        kubectl_output = subprocess.run(
+            kubectl_cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            check=True
+        )
+    except subprocess.CalledProcessError as e:
+        return {"error": f"Error fetching site-init secret: {e.stderr}"}
```
mharding-hpe: The docstring says this function returns a bool, but this is a dict

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,286 @@
...
+    except Exception as e:
+        return {"error": str(e)}
```
mharding-hpe: The docstring says this function returns a bool, but this is a dict

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,286 @@
...
+    # Parse JSON output
+    secret_data = json.loads(kubectl_output.stdout)
+
+    # Extract and decode the base64 data
+    try:
+        encoded_yaml = secret_data["data"]["customizations.yaml"]
+        decoded_yaml = base64.b64decode(encoded_yaml).decode("utf-8")
+        data = yaml.safe_load(decoded_yaml)
+    except Exception as e:
+        return {"error": f"Failed to decode or parse YAML: {str(e)}"}
```
mharding-hpe: The docstring says this function returns a bool, but this is a dict

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,288 @@
...
+    Main function to execute the rollout restart of critical services.
+    """
+    if not rr_enabled():
+        # RR is not enabled
+        if is_cluster_policy_applied("insert-labels-topology-constraints"):
+            # RR is not enabled, but the cluster policy is applied -- this is an error
+            err_exit("Rack Resiliency is not enabled  but ClusterPolicy 'insert-labels-topology-constraints' is applied. Skipping restart.")
+        ## RR is not enabled and the cluster policy is not applied
+        print("Rack Resiliency is not enabled and ClusterPolicy 'insert-labels-topology-constraints' is not applied.Skipping restart.")
+        sys.exit(1)
```
mharding-hpe: Should this be an error exit? This seems like a normal case -- RR is not enabled and the policy is not applied.

(obviously this may change after the changes discussed on the call are made)

upgrade/scripts/k8s/upgrade_k8s_steps.sh
```diff
@@ -119,5 +119,18 @@ else
...
+  echo "INFO Rack Resiliency Critical services rollout restart completed, skipping."
+else
+  # For each CS in static ConfigMap, perform rollout restart of the critical service
+  python3 rr_critical_service_restart.py
+  if [[ $? -eq 0 ]]; then
+    echo "INFO Successfully completed critical services rollout restart"
+    touch "$DONE_DIR/rr_cs_rollout_restart.done"
+  else
+    echo "ERROR Either rollout restart of the critical services is not performed or failed"
```
mharding-hpe: Doesn't this need to have an `exit 1` like the other error clauses earlier in this script, so that this causes the failure to be visible to IUF?

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,291 @@
...
+            continue
+
+        # Restart the service
+        restart_command = ["kubectl", "rollout", "restart", f"{resource_type}/{name}", "-n", namespace]
+        try:
+            subprocess.run(restart_command, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
+        except subprocess.CalledProcessError as e:
+            if "not found" in e.stderr.lower():
+                print(f"Skipping {resource_type}/{name}: resource not found in namespace {namespace}")
+                continue
```
mharding-hpe: Don't need this now that the `get` command is being called inside its own `try`/`except` block
```suggestion
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,288 @@
...
+    try:
+        kubectl_output = subprocess.run(
+            kubectl_cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            check=True
+        )
+    except subprocess.CalledProcessError as e:
+        print(f"Error fetching site-init secret: {e.stderr}")
```
mharding-hpe:
```suggestion
        print_err(f"Error fetching site-init secret: {e.stderr}")
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,288 @@
...
+        return False
+    except Exception as e:
+        print("error": str(e))
```
mharding-hpe:
```suggestion
        print_err(f"error: {e}")
```

upgrade/scripts/k8s/rr_critical_service_restart.py
```diff
@@ -0,0 +1,288 @@
...
+    # Parse JSON output
+    secret_data = json.loads(kubectl_output.stdout)
+
+    # Extract and decode the base64 data
+    try:
+        encoded_yaml = secret_data["data"]["customizations.yaml"]
+        decoded_yaml = base64.b64decode(encoded_yaml).decode("utf-8")
+        data = yaml.safe_load(decoded_yaml)
+    except Exception as e:
+        print(f"Failed to decode or parse customizations YAML: {str(e)}")
```
mharding-hpe:
```suggestion
        print_err(f"Failed to decode or parse customizations.yaml: {e}")
```

upgrade/scripts/k8s/upgrade_k8s_steps.sh
```diff
@@ -119,5 +119,18 @@ else
...
+  echo "INFO Rack Resiliency Critical services rollout restart completed, skipping."
+else
+  # For each CS in static ConfigMap, perform rollout restart of the critical service
+  python3 rr_critical_service_restart.py
+  if [[ $? -eq 0 ]]; then
+    echo "INFO Critical Services rollout restart not performed or successfully completed."
+    touch "$DONE_DIR/rr_cs_rollout_restart.done"
+  else
+    echo "ERROR Critical Services rollout restart failed."
+  fi
```
mharding-hpe:
```suggestion
  python3 rr_critical_service_restart.py || {
    echo "ERROR Critical Services rollout restart failed."
    exit 1
  }
  echo "INFO Critical Services rollout restart successfully completed, if it was necessary"
  touch "$DONE_DIR/rr_cs_rollout_restart.done" || {
    echo "ERROR Failed to create done file for Rack Resiliency critical services rollout restart"
    exit 1
  }
```

upgrade/scripts/k8s/upgrade_k8s_steps.sh
```diff
@@ -119,5 +119,20 @@ else
   fi
 fi
 
+if [[ -f "$DONE_DIR/rr_cs_rollout_restart.done" ]]; then
+  echo "INFO Rack Resiliency Critical services rollout restart completed, skipping."
```
mharding-hpe:
```suggestion
  echo "INFO Rack Resiliency Critical services rollout restart already completed, skipping."
```

upgrade/scripts/k8s/upgrade_k8s_steps.sh
ravikanth-nalla-hpe: Can you check and resolve the comment?

mharding-hpe (changes requested):
The current changes look good, but based on our discussion during the call earlier, there is more required:

1. Move rack resiliency to be one of the services that is updated early in `prerequisites.sh`
2. Move the step to enable it during upgrades from its current location to one of the pre-upgrade prep steps

upgrade/scripts/k8s/upgrade_k8s_steps.sh
mharding-hpe: It hadn't been resolved, but I fixed it

mharding-hpe (approved):
This should not merge until the changes to the service itself are made:
https://github.com/Cray-HPE/cray-rrs/pull/15
https://github.com/Cray-HPE/cray-rrs/pull/16

rajeshranganathan-hpe (approved):

operations/iuf/workflows/product_delivery.md
```diff
@@ -2,14 +2,107 @@
 
 This section ensures the product content is loaded onto the system and available for later steps in the workflow.
 
-1. [Execute the IUF `process-media` and `pre-install-check` stages](#1-execute-the-iuf-process-media-and-pre-install-check-stages)
-1. [Update `customizations.yaml`](#2-update-customizationsyaml)
-1. [Populate admin directory with files defining site preference](#3-populate-admin-directory-with-files-defining-site-preferences)
-1. [Execute the IUF `deliver-product` stage](#4-execute-the-iuf-deliver-product-stage)
-1. [Perform manual product delivery operations](#5-perform-manual-product-delivery-operations)
-1. [Next steps](#6-next-steps)
+1. [Configure optional CSM features](#1-configure-optional-csm-features)
```
mharding-hpe, line 5:
```suggestion
1. [Configure optional CSM features](#1-configure-optional-csm-features)
  - [Rack Resiliency](#rack-resiliency)
```

operations/iuf/workflows/product_delivery.md
```diff
@@ -2,14 +2,108 @@
...
+  - [Rack Resiliency](#rack-resiliency)
```
mharding-hpe:
```suggestion
    - [Rack Resiliency](#rack-resiliency)
```

PR #6252, opened 2025-08-13
CASMPET-7649: Restore iSCSI selective personalization feature; update DNS update procedure

mharding-hpe:
This restores the removed iSCSI selective node personalization documentation, now that we have resolved the problem that caused its removal. That covers the only part of this PR which is not a simple revert -- the iSCSI procedures have been updated to reflect the fact that any time the DNS records need to be updated, re-running the iSCSI Ansible will address it.

The bulk of this PR (the restored content) already went through the review and approval process. It wasn't removed because it had problems, only because it needed one detail added, which this PR adds.

No backports -- this is relevant only for CSM 1.7


operations/iscsi_sbps/iSCSI_SBPS_Verification.md
```diff
@@ -33,8 +33,12 @@ iSCSI configuration. During [Management Node Personalization](../configuration_m
 this playbook does the following:
 
 * Installs the [SBPS Marshal Agent](#sbps-marshal-agent)
-* Adds Kubernetes labels to all worker NCNs, designating them as iSCSI-enabled targets
-* Creates DNS SRV and A records for the workers
+* Adds Kubernetes labels to the worker NCNs that are designated to be iSCSI targets
+* Removes the labels from workers that are not designated to be iSCSI targets
+* Creates DNS SRV and A records for the workers.
```
mharding-hpe:
```suggestion
* Updates the DNS SRV and A records for the workers so that they match the set of workers that are enabled as iSCSI targets.
```

aasha-hpe (approved):

ravikanth-nalla-hpe (approved):
Looks good to me.

PR #6261, opened 2025-08-19
CASM-5705: Need to document (@csm-docs) steps on enabling Rack Resiliency post install/upgrade

keshav-06-hpe:
# Description

Adding procedure to enable and setup Rack Resiliency post install or upgrade.

Resolves:
[CASM-5705](https://jira-pro.it.hpe.com:8443/browse/CASM-5705)

# Checklist


rajeshranganathan-hpe (approved):

arka-pramanik-hpe (approved):

ravikanth-nalla-hpe (approved):
Looks good to me.

operations/rack_resiliency/Enabling_Rack_Resiliency.md
```diff
@@ -5,9 +5,9 @@ Enabling Rack Resiliency is the first step for setting up Rack Resiliency.
 **NOTE:**
 
 * By default, the Rack Resiliency feature is disabled.
-* Rack Resiliency can be enabled only during fresh install of CSM 1.7 or an
-  upgrade from CSM 1.6 to CSM 1.7.
-* Rack Resiliency cannot be disabled after it has been enabled during the install or upgrade.
+* Rack Resiliency can be enabled during fresh install of CSM 1.7 or an
+  upgrade from CSM 1.6 to CSM 1.7. Also, it can be enabled anytime after an install or upgrade.
```
mharding-hpe:
```suggestion
  upgrade from CSM 1.6 to CSM 1.7. It can also be enabled as an operational procedure on a
  system running CSM 1.7+, outside of an install or upgrade.
```

mharding-hpe (changes requested):
I am still reviewing the content of this PR, but I have a higher-level concern that this does not make sense to pull into CSM 1.7.0, based on the inclusion criteria we discussed yesterday. Apart from the doc review itself, we shouldn't merge this until we get clarity on that.

mharding-hpe (changes requested):
I have done a quick linting pass, but there is quite a bit more to be done on that front.

I've also seen several problems with the procedures that will need to be corrected.

But before I spend a lot of time now doing more linting or giving in depth feedback on the procedural issues, first we should determine whether this is going to be included in CSM 1.7.0. Because if not, I have higher priority work to do, and this can be tabled until later.

mharding-hpe:
I made a separate PR for the CASM-5709 changes, so that can get merged without being delayed by this PR
https://github.com/Cray-HPE/docs-csm/pull/6264

mharding-hpe:
The latest discussion with release management indicates this will not merge prior to CSM 1.7.0 being released.

I'm going to move this PR into draft mode. As I have time, I will try to clean up the linting issues, and I will leave feedback on the problems with the procedure that I see. After the procedure is fixed, and after the 1.7.0 release has shipped, then the updated procedure can be tested, and based on that, we can figure out when this PR should merge, and whether the procedure should support CSM 1.7.0 systems, or if we're going to require them to be at CSM 1.7.1.

This will also give time to provide the procedure to disable RR as well, if possible, since that will certainly be something else customers will want to be able to do.

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
```diff
@@ -0,0 +1,293 @@
...
+    ```
+
+    Example output (in a case where only the Ceph zone prefix was set):
+
+    ```yaml
+    enabled: true
+    ceph_zone_prefix: my-ceph-prefix
+    ```
+
+## 2. Run Ansible plays
```
mharding-hpe, line 91: Following the procedures in this section will only run the Ansible plays on one master NCN and one storage NCN, but they should be run on all of them. 

In addition, this procedure will cause those nodes to re-run the entire configuration, not just the RR layer. Even for a node that has previously been configured, this can take a while. 

I think you could address both of these issues, and simplify this procedure considerably, by instead doing something like we did for CASMPET-7628. For that ticket, [this script](https://github.com/Cray-HPE/docs-csm/blob/release%2F1.7/scripts%2Foperations%2Fconfiguration%2Frefresh_worker_iscsi_config.py) was created to force just the iSCSI playbook to re-run on worker nodes. You could do something similar here

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
```diff
@@ -0,0 +1,293 @@
...
+    [Troubleshoot CFS Issues](../configuration_management/Troubleshoot_CFS_Issues.md).
+
+## 3. Deploy Helm chart
+
+Deploy [`cray-rrs`](cray-rrs_Deployment.md) in `rack-resiliency` namespace using the following procedure:
+
+1. (`ncn-mw#`) Get current CSM installation location.
+
+    ```bash
+    . /etc/cray/upgrade/csm/myenv
```
mharding-hpe: This file will not exist on a fresh install, only on an upgraded system

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
```diff
@@ -0,0 +1,293 @@
...
+    echo $CSM_ARTI_DIR
+    ```
+
+    Example output:
+
+    ```text
+    /etc/cray/upgrade/csm/test-activity/2148565/media/csm-1.7.0-rc.2
```
mharding-hpe: Even on an upgraded system, there is no guarantee that the customer will still have the media directory populated with these artifacts.

The RR chart should already be deployed, since it's deployed on all 1.7 systems. And all of the CSM charts are in Nexus. So I think you would be better off figuring out the currently deployed chart version (using helm/loftsman, probably), and then just redeploying it from Nexus. 

Alternatively, this seems like overkill if all you are trying to do is undo the post-hook changes that removed the `exclude` from the Kyverno policy. So an alternative would be to just extract the post-install/upgrade hook and execute just that.

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
```diff
@@ -0,0 +1,293 @@
...
+    ```
+
+1. (`ncn-mw#`) Install the helm chart on the cluster.
```
mharding-hpe: This says install, but as previously noted, it is already going to be installed.

mharding-hpe (changes requested):

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
keshav-06-hpe: Taken Care

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
keshav-06-hpe: Since RR chart is already deployed so this is not needed
Taken care.

operations/rack_resiliency/Enable_Rack_Resiliency_on_a_Running_System.md
keshav-06-hpe: Since RR chart is already deployed so this is not needed
Taken care.

PR #6178, opened 2025-08-01
CASMSEC-585: Added a known issue regarding kyverno admission reports

dineshfomrdj:
# Description
In CSM 1.6.x, multiple admission reports are being generated that might result in the downtime of etcs. A workaround was provided in the doc to address that issue

Relates to:
- [CASMSEC-585](https://jira-pro.it.hpe.com:8443/browse/CASMSEC-585)
# Checklist

- [ ] If I added any command snippets, the steps they belong to follow the prompt conventions (see [example][1]).
- [ ] If I added a new directory, I also updated `.github/CODEOWNERS` with the corresponding team in [Cray-HPE][2].
- [x] My commits or Pull-Request Title contain my JIRA information, or I do not have a JIRA.

 
[1]: https://github.com/Cray-HPE/docs-csm/blob/main/introduction/documentation_conventions.md#using-prompts
[2]: https://github.com/Cray-HPE/teams

PR #6304, opened 2025-09-03
CASMSEC-585: Added a known issue in Kyverno reagrding admission reports

dineshfomrdj:
# Description
In CSM 1.6.x, multiple admission reports are being generated that might result in the downtime of etcs. A workaround was provided in the doc to address that issue

Relates to:
- [CASMSEC-585](https://jira-pro.it.hpe.com:8443/browse/CASMSEC-585)
# Checklist


operations/kubernetes/Kyverno.md
```diff
@@ -456,3 +456,20 @@ and [match/exclude](https://release-1-10-0.kyverno.io/docs/writing-policies/matc
 
     If the issue still persists after restarting the Kyverno pods, then HPE recommends disabling the background scanning in the Kyverno policy.
     Set `background` to `false` to disable background scanning in the Kyverno policy.
+* Multiple Kyverno admission reports may cause etcd downtime during upgrade to CSM 1.6.
```
mharding-hpe, line 459: This states that the problem *may* happen. I think an administrator reading this would reasonably wonder if this means that they should proactively do this procedure to avoid the problem.

The text doesn't convey whether this is a common problem (where it would then make sense to do this proactively) or if it is a rare problem (where it would then perhaps make sense to only worry about this if they hit it).

(From comment on original PR: https://github.com/Cray-HPE/docs-csm/pull/6178#discussion_r2248267925)

mharding-hpe, line 459: This should describe the symptoms an admin would observe during an upgrade that would indicate this problem is happening. 

Specifically, it should explain:
1. How the etcd downtime will manifest itself during the upgrade
2. If there is a way to determine that the etcd downtime is caused by this versus something else

mharding-hpe, line 459: At what point during an upgrade can this happen? I assume there must be something in the upgrade which causes this, so we should at least be able to say that it would start being observed (for example) after CSM services are upgraded .

mharding-hpe (changes requested):

operations/kubernetes/Kyverno.md
dineshfomrdj: Added this as an additional note at the end

dineshfomrdj: Added the description on how to check the symptoms

dineshfomrdj: Added that info in the latest updated document

troubleshooting/README.md
```diff
@@ -91,6 +91,7 @@ to the exiting problem seen into the existing search. (The example searches for
 * [Soft Deleted IMS Image Metadata is Lost](known_issues/IMS_Soft_Deleted_Image_Metadata_is_Lost.md)
 * [IMS Created Image Stores Incorrect Metadata](known_issues/IMS_Created_Image_Stores_Incorrect_Metadata.md)
 * [PostgreSQL System ID Mismatch](known_issues/postgres_system_id_mismatch.md)
+* [Kyverno policy management](../operations/kubernetes/Kyverno.md#known-issues)
```
mharding-hpe:
```suggestion
* [Kyverno policy management known issues](../operations/kubernetes/Kyverno.md#known-issues)
```

mharding-hpe (approved):

mharding-hpe:
@dineshfomrdj Can this problem also happen during upgrades from CSM 1.6 to CSM 1.7? Or is it only with upgrades from CSM 1.5 to CSM 1.6?

If it can possibly also happen during upgrades from CSM 1.6 to CSM 1.7, then a PR is also needed to add this to the known issues in the `release/1.7` branch.

dineshfomrdj:
> @dineshfomrdj Can this problem also happen during upgrades from CSM 1.6 to CSM 1.7? Or is it only with upgrades from CSM 1.5 to CSM 1.6?
> 
> If it can possibly also happen during upgrades from CSM 1.6 to CSM 1.7, then a PR is also needed to add this to the known issues in the `release/1.7` branch.

No, it's only an issue during upgrades from CSM 1.5 to CSM 1.6

yarlagadda-srinivasa-rao (approved):
It looks fine for me.

yarlagadda-srinivasa-rao:
This is happening during 1.5 to 1.6 upgrades.

PR #6508, opened 2026-02-12
CASMINST-7541: Linting

mharding-hpe:
Just linting of issues I noticed while doing other docs work. No real content changes.

Backports:

* 1.6: https://github.com/Cray-HPE/docs-csm/pull/6509
* 1.5: https://github.com/Cray-HPE/docs-csm/pull/6510
* 1.4: https://github.com/Cray-HPE/docs-csm/pull/6511
* 1.3: https://github.com/Cray-HPE/docs-csm/pull/6512
* 1.2: https://github.com/Cray-HPE/docs-csm/pull/6513
* 1.0: https://github.com/Cray-HPE/docs-csm/pull/6514

PR #6514, opened 2026-02-12
CASMINST-7541: Linting

mharding-hpe:
Backport of https://github.com/Cray-HPE/docs-csm/pull/6508

### DIFF
--- a/operations/utility_storage/Ceph_Orchestrator_Usage.md
+++ b/operations/utility_storage/Ceph_Orchestrator_Usage.md
@@ -318,113 +318,113 @@ ncn-s# ceph orch host ok-to-stop <hostname>
 Remove a host
 
 ```bash
-ncn-s# cephorch host rm <hostname>
+ncn-s# ceph orch host rm <hostname>
 ```
 
 Update a host address
 
 ```bash
-ncn-s# cephorch host set-addr <hostname> <addr>
+ncn-s# ceph orch host set-addr <hostname> <addr>
 ```
 
 List services known to orchestrator
 
 ```bash
-ncn-s# cephorch ls [<service_type>] [<service_name>] [--export] [plain|json|json-pretty|yaml] [--refresh]
+ncn-s# ceph orch ls [<service_type>] [<service_name>] [--export] [plain|json|json-pretty|yaml] [--refresh]
 ```
 
 Remove OSD services
 
 ```bash
-ncn-s# cephorch osd rm <svc_id>... [--replace] [--force]
+ncn-s# ceph orch osd rm <svc_id>... [--replace] [--force]
 ```
 
 Status of OSD removal operation
 
 ```bash
-ncn-s# cephorch osd rm status [plain|json|json-pretty|yaml]
+ncn-s# ceph orch osd rm status [plain|json|json-pretty|yaml]
 ```
 
 Remove OSD services
 
 ```bash
-ncn-s# cephorch osd rm stop <svc_id>...
+ncn-s# ceph orch osd rm stop <svc_id>...
 ```
 
 Pause orchestrator background work
 
 ```bash
-ncn-s# cephorch pause
+ncn-s# ceph orch pause
 ```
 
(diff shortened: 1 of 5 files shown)


operations/utility_storage/Ceph_Orchestrator_Usage.md
```diff
@@ -154,11 +171,11 @@ ncn-s003  /dev/vdc  hdd   4797e919-667e-4376-b   107G  Unknown  N/A    N/A    No
 ncn-s003  /dev/vdd  hdd   3b2c090d-37a0-403b-a   107G  Unknown  N/A    N/A    No
 ```
 
-**`IMPORTANT:`** If you use the `--wide`, it will give the reasons a drive is not `Available.`. This **DOES NOT** mean something is wrong. If Ceph already has the drive provisioned, then you may see similar reasons
+**`IMPORTANT:`** If `--wide` is used, it will give the reasons a drive is not `Available.`. This **DOES NOT** mean something is wrong. If Ceph already has the drive provisioned, then one may see similar reasons
```
jpdavis-prof: Missing period at end of sentence.

```suggestion
**`IMPORTANT:`** If `--wide` is used, it will give the reasons a drive is not `Available.`. This **DOES NOT** mean something is wrong. If Ceph already has the drive provisioned, then one may see similar reasons.
```

operations/utility_storage/Ceph_Orchestrator_Usage.md
```diff
@@ -362,7 +381,7 @@ Select orchestrator module backend
 ncn-s# cephorch set backend <module_name>
 ```
 
-Start, stop, restart, redeploy, or reconfig an entire service (i.e. all daemons)
+Start, stop, restart, redeploy, or reconfigure an entire service (i.e. all daemons):
```
jpdavis-prof, line 385: Typo: "cephorch" should be "ceph orch" (missing space).

```suggestion
ncn-s# ceph orch set backend <module_name>
```

operations/utility_storage/Ceph_Orchestrator_Usage.md
```diff
@@ -242,13 +261,13 @@ Redeploy a daemon (with a specific image)
...
 
 ```bash
 ncn-s# ceph orch daemon rm <names>... [--force]
 ```
 
-Start, stop, restart, (redeploy,) or reconfig a specific daemon
+Start, stop, restart, or reconfigure a specific daemon:
 
 ```bash
 ncn-s# ceph orch daemon start|stop|restart|reconfig <name>
```
jpdavis-prof, line 273: Typo: "cephorch" should be "ceph orch" (missing space).

```suggestion
ncn-s# ceph orch start|stop|restart|redeploy|reconfig <service_name>
```

jpdavis-prof:
Looks like `cephorch` appears quite a few times.

jpdavis-prof (approved):
All spelling/grammar suggestions have been applied. Looks good.

operations/utility_storage/Ceph_Orchestrator_Usage.md
mharding-hpe: I am not sure I understand this suggestion. Your comment mentions a missing space, but I don't see any missing space. And your suggestion looks like it is adding a command outside of a backtick-enclosed box. Can you look at this one again?

operations/utility_storage/Ceph_Orchestrator_Usage.md
mharding-hpe: Same thing here. I am not sure I understand this suggestion. Your comment mentions a missing space, but I don't see any missing space. And your suggestion looks like it is changing other parts of the command. Can you take another look and clarify?

mharding-hpe:
> Looks like `cephorch` appears quite a few times.

Yeah, I see that (although not in the two cases where you made suggestions for it).

Are you confident that this is not some difference with the old Ceph version versus the newer ones? If so, I'm happy to fix all of those.

